PR #3, opened 2021-11-29
Update README.md

vilmibm:
- add note about beta
- explain the two general ways to use `gh` stuff
- add another example showing an alternative to `Exec`

An open question: I'd like to be able to point at reference docs. are we generating those?

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,6 +55,10 @@ func main() {
 
 See [examples][examples] for more use cases.
 
+## Reference Documentation
+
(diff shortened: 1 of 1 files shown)


samcoe (approved):
LGTM! Thanks for the clean up. Reference docs are https://pkg.go.dev/github.com/cli/go-gh and are automatically generated by `pkg.go.dev`.

kdcloverkid:
> - add note about beta
> - explain the two general ways to use `gh` stuff
> - add another example showing an alternative to `Exec`
> 
> An open question: I'd like to be able to point at reference docs. are we generating those?

5

Issue #4, opened 2021-12-03
Status code extraction on response/error


sridharavinash:
I had a quick question/feedback - I was wondering if a status code could be included along with the response from the gh.RESTClient().Get? 

for example I was making a call
```go
client, err := gh.RESTClient(options)
<snip>
err = client.Get("enterprises/avocado-corp/audit-log", &response)
Unfortunately err  cannot be type-casted to api.httperror since api.httperror is not exported, so I’d have to rely on string parsing to figure out the error code here. Alternatively api.httpError could be exported to api.HTTPError , in which case the caller can then typecast the error and get all the relevant bits in the struct and do things like retry on certain statusCodes etc.
e.g
err = client.Get(blah)
if err != nil {
    httpError := err.(*api.HttpError)
    if httpError.statusCode == 502 {
            ...retry
     }else{
            ...do something else
      }
}
```
or the other option would be to read the statusCode, if it’s set on the response.

Does that make sense?

I started a PR to export api.HttpError here - https://github.com/cli/go-gh/commit/6596865f9dfc7df98da40b6c74fe1b3e918e8d4c 

Thanks for reading ❤️ .

samcoe:
Closed in https://github.com/cli/go-gh/pull/6, this will be available in the next release.

PR #6, opened 2021-12-03
Export api.HTTPError

sridharavinash:
This is an attempt to fix the issue mentioned in #4. 

With the change in here, `api.HTTPError` would be exported and it would allow the caller to the REST API client to do things like 
```go
client, _ := gh.RESTClient(options)
err := client.Get(blah)
if err != nil {
    httpError := err.(*api.HTTPError)
    if httpError.StatusCode == 502 {
            ...retry
     }else{
            ...do something else
      }
}
```

Thanks for the 👀 💖 .

### DIFF
--- a/internal/api/http.go
+++ b/internal/api/http.go
@@ -5,7 +5,6 @@ import (
 	"fmt"
 	"io"
 	"net/http"
-	"net/url"
 	"os"
 	"path/filepath"
 	"regexp"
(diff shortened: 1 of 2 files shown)


samcoe:
@sridharavinash Thanks for opening this! I made one commit to move the `HTTPError` out of the internal directory, my understanding is that Go automatically keeps anything in that directory from being exported outside the package so we needed to move the code out of there also.

samcoe (approved):

PR #11, opened 2022-01-08
go get should not contain URL scheme

devonhk:
go get fails with example code:
```
go get https://github.com/cli/go-gh
go get: malformed module path "https:/github.com/cli/go-gh": invalid char ':'
```


samcoe (approved):
Nice catch! Thanks.

PR #12, opened 2022-01-10
Add support for GH_HOST and expose repository.Parse

### LINKED ISSUE #8: Add support for GH_REPO environment variable
The `CurrentRepository` function should take into account the `GH_REPO` environment variable, and if it is specified return that repository rather than the current flow of looking at the filesystem and git remotes. 

As part of this work we should make the parsing function used for parsing `GH_REPO` into a `repository` available for user consumption.

cc https://github.com/cli/go-gh/issues/5

samcoe:
This PR adds support for the `GH_HOME` environment variable to the `CurrentRepository` function. To support this new functionality I added a new function `Parse` to the new `repository` package. The `Parse` function is exported as the operation of parsing a string into its repository information seems like a useful utility for extension authors to have.   
 
Closes https://github.com/cli/go-gh/issues/8

### DIFF
--- a/internal/git/url.go
+++ b/internal/git/url.go
@@ -6,7 +6,7 @@ import (
 	"strings"
 )
 
-func isURL(u string) bool {
+func IsURL(u string) bool {
 	return strings.HasPrefix(u, "git@") || isSupportedProtocol(u)
 }
 
@@ -15,6 +15,7 @@ func isSupportedProtocol(u string) bool {
 		strings.HasPrefix(u, "git+ssh:") ||
 		strings.HasPrefix(u, "git:") ||
 		strings.HasPrefix(u, "http:") ||
+		strings.HasPrefix(u, "git+https:") ||
 		strings.HasPrefix(u, "https:")
 }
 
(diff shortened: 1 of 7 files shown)


internal/git/url.go
```diff
@@ -60,7 +65,7 @@ func parseURL(rawURL string) (u *url.URL, err error) {
 }
 
 // Extract GitHub repository information from a git remote URL.
-func repoInfoFromURL(u *url.URL) (host string, owner string, repo string, err error) {
+func RepoInfoFromURL(u *url.URL) (host string, owner string, name string, err error) {
```
mislav, line 68: This function has 4 return values, which makes it a poor public API in my view.

A more useful function than the current `ParseURL` + `RepoInfoFromURL` might be a single method that parses a string URL into `(Repository, error)` return values. What do you think?

mislav (changes requested):
This looks great; I'm just wondering if we could improve the API for parsing git remote URLs?

internal/git/url.go
samcoe: Agreed, this and `ParseURL` are the internal helper functions for the `Parse` function defined below in `pkg/repository/repository.go` and are not exposed to end users of the package.

mislav: Ahh, completely missed the fact that this was in an internal package 🤦

mislav (approved):

PR #15, opened 2022-01-18
Expose a http.Client instance

### LINKED ISSUE #13: Consider exposing a raw http.Client
Hey - this is a great project and helped me get up and running quickly!

I was looking at integrating https://github.com/google/go-github in to my project. It allows you to pass a pre-configured http client when creating a new GitHubClient instance. The abstraction (`RESTClient`) doesn't implement the interface so cannot be used as a parameter.

Would you consider exposing a configured raw http.Client instance or the config package to allow us to build our own?

I'd be happy to help out with the implementation if it's something of use.

chelnak:
Closes #13 

The idea behind this PR is that a developer can opt to configure another library (e.g go-github) with a http client instance that has been configured to use gh-cli auth.


example/httpclient/main.go
```diff
@@ -0,0 +1,32 @@
...
+// that can be returned by the gh.HTTPClient function.
+package main
+
+import (
+	"context"
+	"fmt"
+	"log"
+
+	"github.com/cli/go-gh"
+	"github.com/google/go-github/v42/github"
```
mislav: Let's not add this example because it adds a whole new library to our `go.mod` and it's entirely unnecessary. Go developers typically don't need an example of what to do with a `http.Client` instance.

gh.go
```diff
@@ -107,6 +108,35 @@ func GQLClient(opts *api.ClientOptions) (api.GQLClient, error) {
...
+	}
+	if opts.Host == "" {
+		opts.Host = cfg.Host()
+	}
+	if opts.AuthToken == "" {
+		token, err = cfg.AuthToken(opts.Host)
+		if err != nil {
+			return nil, err
+		}
+		opts.AuthToken = token
```
mislav: This `http.Client` instance will be configured with a `github.com` authentication token. However, a `http.Client` instance can be used to make arbitrary requests to any host, and we don't know how it will be used. We absolutely would *not* want that the `github.com` auth token is sent to non-`github.com` hosts. https://github.com/cli/go-gh/blob/6722e5635158d5b607f7d955253e826f7c87b5a4/internal/api/http.go#L219

If we were to expose a generic `http.Client`, we'd have to take precaution so that the token isn't set to other hosts. It might be as simple as only sending the authentication token with requests to hostnames that match `opts.Host`. What do you think?

mislav (changes requested):

example/httpclient/main.go
chelnak: 👍

gh.go
chelnak: Yeah this sounds sensible to me. I'll make a start by adding something that stops the auth token being forward to non `opts.Host` requests as you suggest.

Thanks!

chelnak: @mislav I've made a small change that stops the authorization header from being included if the hosts don't match.

The interesting thing was that the configured host and request host will never be an explicit match (github.com vs api.github.com) so I am checking for existence of the one in the other.

Unsure if this is a viable solution so your feedback is very welcome!

chelnak: I've added some tests for the above change.  👍

chelnak:
~I'm also wondering if I should move `NewHttpClient` from `http_client.go` in to `http.go`... it's not an interface with methods or anything so feel like it's a bit overkill having it in a new file.~

Resolved by exporting the internal method.

internal/api/http_client.go
```diff
@@ -0,0 +1,12 @@
+package api
+
+import (
+	"net/http"
+
+	"github.com/cli/go-gh/pkg/api"
+)
+
+func NewHTTPClient(opts *api.ClientOptions) *http.Client {
+	c := newHTTPClient(opts)
```
samcoe: Seems like we could just export `newHTTPClient` rather than add this function.

gh.go
```diff
@@ -107,6 +108,35 @@ func GQLClient(opts *api.ClientOptions) (api.GQLClient, error) {
 	return iapi.NewGQLClient(opts.Host, opts), nil
 }
 
+// HTTPClient builds a client that can be passed to another library.
+// As part of the configuration a hostname, auth token, and default set of headers are resolved
+// from the gh environment configuration. These behaviors can be overridden using the opts argument.
```
samcoe, line 77: We should add some comments about how providing the host option here will not actually direct a request to that host which is different from the other clients. Also, we should add some comments about how the auth token is protected from being sent to arbitrary hosts.

internal/api/http.go
```diff
@@ -232,12 +237,20 @@ func newHeaderRoundTripper(authToken string, headers map[string]string, rt http.
...
 		headers[accept] = a
 	}
-	return headerRoundTripper{headers: headers, rt: rt}
+	return headerRoundTripper{host: host, headers: headers, rt: rt}
 }
 
 func (hrt headerRoundTripper) RoundTrip(req *http.Request) (*http.Response, error) {
+	// If request is not for the host we have configured, do not add authorization headers.
+	if !isValidHost(req.URL.Host, hrt.host) && hrt.headers[authorization] != "" {
+		delete(hrt.headers, "Authorization")
```
samcoe:
```suggestion
		delete(hrt.headers, authorization)
```

chelnak: The key in this map is actually `Authorization`. Applying your suggestion would make the test fail.

map:

```go
map[Accept:application/vnd.github.merge-info-preview+json, application/vnd.github.nebula-preview, application/vnd.github.antiope-preview, application/vnd.github.shadow-cat-preview Authorization:token oauth_token Content-Type:application/json; charset=utf-8 Time-Zone:Europe/London User-Agent:go-gh]
```

failing test:

```go
--- FAIL: TestNewHTTPClientWithDifferentHost (0.00s)
    --- FAIL: TestNewHTTPClientWithDifferentHost/removes_authorization_header_for_a_different_host (0.00s)
        /Users/craigg/code/go-gh/internal/api/http_test.go:145: 
            	Error Trace:	http_test.go:145
            	Error:      	Not equal: 
            	            	expected: ""
            	            	actual  : "token oauth_token"
            	            	
            	            	Diff:
            	            	--- Expected
            	            	+++ Actual
            	            	@@ -1 +1 @@
            	            	-
            	            	+token oauth_token
            	Test:       	TestNewHTTPClientWithDifferentHost/removes_authorization_header_for_a_different_host
FAIL
FAIL	github.com/cli/go-gh/internal/api	0.141s
FAIL
```

internal/api/http_client.go
chelnak: No problemo, will get this done shortly.

gh.go
chelnak: Sounds sensible to me. Commit incoming.

internal/api/http.go
chelnak: Ignore that. Completely overlooked that it was a const 😸

internal/api/http.go
```diff
@@ -232,12 +237,20 @@ func newHeaderRoundTripper(authToken string, headers map[string]string, rt http.
...
 		headers[accept] = a
 	}
-	return headerRoundTripper{headers: headers, rt: rt}
+	return headerRoundTripper{host: host, headers: headers, rt: rt}
 }
 
 func (hrt headerRoundTripper) RoundTrip(req *http.Request) (*http.Response, error) {
+	// If request is not for the host we have configured, do not add authorization headers.
+	if !isValidHost(req.URL.Host, hrt.host) && hrt.headers[authorization] != "" {
+		delete(hrt.headers, authorization)
```
mislav: Instead of deleting the `authorization` header for invalid hosts, let's conditionally add it only for valid hosts.

internal/api/http.go
```diff
@@ -232,12 +237,20 @@ func newHeaderRoundTripper(authToken string, headers map[string]string, rt http.
 		a += ", application/vnd.github.shadow-cat-preview"
 		headers[accept] = a
 	}
-	return headerRoundTripper{headers: headers, rt: rt}
+	return headerRoundTripper{host: host, headers: headers, rt: rt}
 }
 
 func (hrt headerRoundTripper) RoundTrip(req *http.Request) (*http.Response, error) {
+	// If request is not for the host we have configured, do not add authorization headers.
+	if !isValidHost(req.URL.Host, hrt.host) && hrt.headers[authorization] != "" {
```
mislav: Please use `u.Hostname()` instead of `u.Host`, since the former strips out the port name if present

internal/api/http.go
```diff
@@ -187,16 +187,21 @@ func inspectableMIMEType(t string) bool {
 	return strings.HasPrefix(t, "text/") || jsonTypeRE.MatchString(t)
 }
 
+func isValidHost(reqHost, hrtHost string) bool {
+	return strings.Contains(reqHost, hrtHost)
```
mislav: Hostnames in URLs are case-insensitive, so you will want to factor that in when performing comparisons.

To compare whether two values are identical hostnames, you can use `strings.EqualFold(host1, host2)`.

To check whether one hostname is a subdomain of the other, enforce the use of the `.` separator and do strict suffix check. For this, `strings.Contains()` is not a good method because it's imprecise: a host called `thegithub.company` does contain the string `github.com` but is *not* a subdomain of `github.com`. Instead, this is a better check: `strings.HasSuffix("api.github.com", ".github.com")`. Again, remember to normalize the case of input arguments before submitting them for the HasSuffix check.

mislav (changes requested):
Getting there!

internal/api/http.go
chelnak: Thanks for this. I have modified as suggested.

internal/api/http.go
chelnak: This is done (i think). I've also reworked the tests to align.

internal/api/http.go
chelnak: I've made this more robust based on your suggestions.

internal/api/http.go
```diff
@@ -187,8 +187,17 @@ func inspectableMIMEType(t string) bool {
...
+	reqHost = strings.ToLower(reqHost)
+	hrtHost = strings.ToLower(hrtHost)
+
+	return (reqHost == hrtHost) ||
+		// // Check if the hrtHost is a subdomain of reqHost
+		// // e.g strings.HasSuffix("api.github.com", "".github.com") == true
+		strings.HasSuffix(reqHost, "."+hrtHost) ||
+		// // Check if the reqHost is a subdomain of hrtHost
+		// // e.g strings.HasSuffix("github.com", "".api.github.com") == false
+		strings.HasSuffix(hrtHost, "."+reqHost)
```
mislav: Let's remove this check; I can't envision a scenario where it would be necessary

internal/api/http.go
```diff
@@ -187,8 +187,17 @@ func inspectableMIMEType(t string) bool {
 	return strings.HasPrefix(t, "text/") || jsonTypeRE.MatchString(t)
 }
 
-func isValidHost(reqHost, hrtHost string) bool {
-	return strings.Contains(reqHost, hrtHost)
+func isRequestHostValid(reqHost, hrtHost string) bool {
```
mislav: Since this function checks whether the request host is under a specific domain (e.g. `github.com`), and not whether a hostname is "valid" (which is vague and unspecified), how about renaming it to `isSameDomain()`?

internal/api/http.go
```diff
@@ -187,8 +187,17 @@ func inspectableMIMEType(t string) bool {
...
+	reqHost = strings.ToLower(reqHost)
+	hrtHost = strings.ToLower(hrtHost)
+
+	return (reqHost == hrtHost) ||
+		// // Check if the hrtHost is a subdomain of reqHost
+		// // e.g strings.HasSuffix("api.github.com", "".github.com") == true
```
mislav: This comment is wrong and full of typos. You can remove it; it should be easy to understand what the logic is doing here.

internal/api/http.go
mislav: Let's avoid acronyms and abbreviations in variable names when not necessary.
- `reqHost` could be `requestHost`
- `hrtHost` (what is "hrt" in this context anyway?) could simply be `domain`

mislav (dismissed):
Thanks for the updates!

internal/api/http.go
chelnak: Whoops - the comments should have been removed. They were for my reference when debugging.

internal/api/http.go
chelnak: Removed

internal/api/http.go
chelnak: Good idea - I've renamed as suggested

chelnak: Yep can see where you are coming from here and have made the suggested changes.

samcoe (approved):
@chelnak This looks great thanks for making the requested changed. I pushed just a couple commits refactoring out a common `resolveOptions` function and adding a bit of 💅

chelnak:
@samcoe Thank you - appreciate the guidance from you both. 😄

chelnak:
Hey! Is there anything more I can do for this PR?

chelnak:
Thank you ❤️ !

Issue #28, opened 2022-04-27
Add httpmock from cli/cli


heaths:
The mock `Registry` and friends from https://github.com/cli/cli/tree/trunk/pkg/httpmock would be very useful for testing extensions.

samcoe:
I like this idea and I think we should do it. I already have migrated it to [go-gh](https://github.com/cli/go-gh/tree/trunk/internal/httpmock) so it really just needs to be moved from the internal folder to export it.

Before making it widely available there are a couple changes I would like to see to the interface while we still can:

1. Should the name change to something like `TransportMock` or `RoundTripperMock`? 
2. Stubs should have names when they are registered with [`Register`](https://github.com/cli/go-gh/blob/c41a1271d97cf42ba0dcf94289da729cdb1554eb/internal/httpmock/registry.go#L22). This will allow for better error messaging directly referencing the stub name.
3. We should remove the more obscure responders that are not used in `go-gh`, for example [`FileResponse`](https://github.com/cli/go-gh/blob/c41a1271d97cf42ba0dcf94289da729cdb1554eb/internal/httpmock/stub.go#L92) and [`ScopesResponder`](https://github.com/cli/go-gh/blob/c41a1271d97cf42ba0dcf94289da729cdb1554eb/internal/httpmock/stub.go#L146). Perhaps there are others there that we can remove.

These will require changes to the `go-gh` test suit so I think approaching it as making the interfaces changes first and then exporting the package in a second PR is the right way to go here.

heaths:
1. I like `TransportMock` a bit better. Seems more obvious and discoverable.
2. 👍
3. While we're at it, would be nice to define a pattern for taking a string vs. object or map to marshal. I had started some changes like this for one of my recent PRs since I already had an object in test I could marshal, and construction of it is more straight forward in some cases. Certainly just a preformatted JSON response string in other cases is great. For HTTP this already exists, but something similar for GraphQL would be nice. Maybe some pattern like `{protocol}{"Request"|"Response"}({"String"|"JSON"})` e.g., "HTTPRequest", "HTTPResponseString", "GraphQLRequest", or "GraphQLResponseJSON".

samcoe:
Great, `TransportMock` it is. 


[...]

PR #29, opened 2022-05-02
Cleanup httpmock

samcoe:
This PR gets the package formerly known as `httpmock` into a state that it can be made public and moved out of the `internal` folder. 

Some major changes:
- Change package name to `transportmock`
- Require names for stubs
- Removed extraneous responders
- Allow responders to handle strings and interface input 

I left comments for open questions inline.

cc https://github.com/cli/go-gh/issues/28


internal/transportmock/matcher.go
```diff
@@ -0,0 +1,59 @@
...
+	"encoding/json"
+	"io"
+	"net/http"
+	"regexp"
+	"strings"
+)
+
+type Matcher func(req *http.Request) bool
+
+func REST(method, p string) Matcher {
```
samcoe, line 14: What do we think about the naming scheme here? Should we prefix/postfix `Match`/`Matcher` to this and the `GQL` function?

internal/transportmock/registry.go
```diff
@@ -0,0 +1,73 @@
...
+	"net/http"
+	"net/http/httputil"
+	"sync"
+	"testing"
+)
+
+type Registry struct {
+	mu       sync.Mutex
+	stubs    []*stub
+	requests []*http.Request
```
samcoe, line 14: I purposefully un-exported these and added a `Requests()` accessor function. It felt smart to move this direction incase we want to change the underlying implementation.

internal/transportmock/registry.go
```diff
@@ -0,0 +1,73 @@
...
+func (r *Registry) RoundTrip(req *http.Request) (*http.Response, error) {
+	var stub *stub
+	r.mu.Lock()
+	for _, s := range r.stubs {
+		if s.matched || !s.matcher(req) {
+			continue
+		}
+		if stub != nil {
+			r.mu.Unlock()
+			dr, _ := httputil.DumpRequestOut(req, true)
```
samcoe, line 57: I think using `DumpRequestOut` leads to nicer looking error messages.

internal/transportmock/responder.go
```diff
@@ -0,0 +1,111 @@
...
+import (
+	"bytes"
+	"encoding/json"
+	"io"
+	"net/http"
+)
+
+type Responder func(req *http.Request) (*http.Response, error)
+
+func HTTPResponse(status int, header *map[string][]string, body interface{}, cb func(*http.Request)) Responder {
```
samcoe, line 12: What do we think about the naming scheme for Responders? Should we prefix/postfix them with `Response`/`Responder`?

internal/transportmock/responder.go
```diff
@@ -0,0 +1,111 @@
...
+			cb(req)
+		}
+		if header == nil {
+			header = &map[string][]string{}
+		}
+		return httpResponse(status, *header, b, req), nil
+	}
+}
+
+func RESTResponse(body interface{}, cb func(map[string]interface{})) Responder {
```
samcoe, line 31: I think having `HTTPResponse` and `RESTResponse` to be a bit confusing. Open to better naming here.

heaths: Any less confusing if you call `RESTResponse` a `JSONResponse`? Afterall, it's just the response itself which happens to be JSON. Perhaps `HTTPResponse` could be `RAWResponse` and uses for non-JSON payloads - perhaps with a content-type? More focus on the payload than how it was retrieved.

samcoe:
Closing as we decided against exporting `httpmock` in favor of migrating to `gock` in https://github.com/cli/go-gh/pull/31.

PR #30, opened 2022-05-04
Add support for http_unix_socket

### LINKED ISSUE #2: Support  gh http_unix_socket configuration option
`gh` supports a `http_unix_socket` option that impacts how API requests get sent, we should take that configuration option into account up the API clients.

samcoe:
This PR adds support for automatically resolving `gh` config `http_unix_socket`. 

Closes https://github.com/cli/go-gh/issues/2

### DIFF
--- a/gh.go
+++ b/gh.go
@@ -16,6 +16,7 @@ import (
 
 	iapi "github.com/cli/go-gh/internal/api"
 	"github.com/cli/go-gh/internal/config"
+	iconfig "github.com/cli/go-gh/internal/config"
 	"github.com/cli/go-gh/internal/git"
 	irepo "github.com/cli/go-gh/internal/repository"
 	"github.com/cli/go-gh/internal/ssh"
@@ -54,8 +55,9 @@ func run(path string, env []string, args ...string) (stdOut, stdErr bytes.Buffer
 }
 
 // RESTClient builds a client to send requests to GitHub REST API endpoints.
-// As part of the configuration a hostname, auth token, and default set of headers are resolved
-// from the gh environment configuration. These behaviors can be overridden using the opts argument.
+// As part of the configuration a hostname, auth token, default set of headers,
+// and unix domain socket are resolved from the gh environment configuration.
+// These behaviors can be overridden using the opts argument.
 func RESTClient(opts *api.ClientOptions) (api.RESTClient, error) {
 	if opts == nil {
 		opts = &api.ClientOptions{}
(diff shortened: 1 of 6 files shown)


internal/api/http.go
```diff
@@ -74,10 +75,15 @@ func NewHTTPClient(opts *api.ClientOptions) http.Client {
...
 
 	transport := http.DefaultTransport
+
 	if opts.Transport != nil {
 		transport = opts.Transport
 	}
 
+	if opts.UnixDomainSocket != "" {
+		transport = newUnixDomainSocketRoundTripper(opts.UnixDomainSocket)
+	}
```
samcoe: Should `opts.Transport` take precedence over the `opts.UnixDomainSocket`? Or should there be a way to define this behavior?

internal/api/http.go
```diff
@@ -99,25 +114,18 @@ func NewHTTPClient(opts *api.ClientOptions) http.Client {
...
-		}
-		c := cache{dir: opts.CacheDir, ttl: opts.CacheTTL}
-		transport = c.RoundTripper(transport)
-	}
+	transport = newHeaderRoundTripper(opts.Host, opts.AuthToken, opts.Headers, transport)
 
 	return http.Client{Transport: transport, Timeout: opts.Timeout}
 }
 
+// TODO: Export function in near future.
```
samcoe, line 122: Handled in https://github.com/cli/go-gh/pull/39

pkg/api/client.go
```diff
@@ -44,8 +44,18 @@ type ClientOptions struct {
...
+	// handle routing to UnixDomainSocket if necessary. Generally, setting Transport
+	// should be reserved for testing purposes.
 	// Default is http.DefaultTransport.
 	Transport http.RoundTripper
+
+	// UnixDomainSocket specifies the Unix domain socket address by which individual
+	// API requests will be routed. If specifed, this will form the base of the API
+	// request transport chain.
+	// Default is no socket address.
+	UnixDomainSocket string
```
samcoe, line 58: Open to better naming suggestions here.

pkg/api/http.go
```diff
@@ -8,11 +8,12 @@ import (
 
 // HTTPError represents an error response from the GitHub API.
 type HTTPError struct {
-	StatusCode  int
-	RequestURL  *url.URL
-	Message     string
-	OAuthScopes string
-	Errors      []HTTPErrorItem
+	AcceptedOAuthScopes string
```
samcoe: Rather than exporting both `AcceptedOAuthScopes` and `OAuthScopes` does it make sense to export all the response headers?

gh.go
```diff
@@ -156,10 +161,19 @@ func resolveOptions(opts *api.ClientOptions, cfg config.Config) error {
 	if opts.AuthToken == "" {
 		token, err = cfg.AuthToken(opts.Host)
 		if err != nil {
-			return err
+			var notFoundError iconfig.NotFoundError
+			if errors.As(err, &notFoundError) {
+				return fmt.Errorf("auth token not found for host %s", opts.Host)
```
mislav: Nit: "authentication token" instead of "auth token"

Also, should this error be of a unique type so that integrators can handle it specifically?

pkg/api/http.go
mislav: Yeah, maybe HTTPError could just expose all headers received from the response! Then integrators (including `cli/cli`) could wrap that

mislav (approved):
Fantastic; thanks for adding this

gh.go
samcoe: I think that is a good idea, but can be implemented in a follow up PR.

PR #31, opened 2022-05-05
Migrate from httpmock to gock

samcoe:
This PR migrates our tests from using our homemade `httpmock` package with the third-party `gock` package. `gock` provides all the functionality that `httpmock` does and more. It also has a nice declarative DSL that is easier to understand, in my opinion. This is an alternative to https://github.com/cli/go-gh/pull/29. 

cc https://github.com/cli/go-gh/issues/28


gh_test.go
```diff
@@ -57,24 +58,24 @@ func TestRESTClient(t *testing.T) {
 	os.Setenv("GH_CONFIG_DIR", tempDir)
 	os.Setenv("GH_TOKEN", "GH_TOKEN")
 
-	http := httpmock.NewRegistry(t)
-	http.Register(
-		httpmock.REST("GET", "some/test/path"),
-		httpmock.StatusStringResponse(200, `{"message": "success"}`),
-	)
+	gock.New("https://api.github.com").
```
mislav, line 62: Instead of `gock.New()`, could we use a gock initialization method that doesn't intercept the default `http.DefaultTransport`, but makes a new transport that we can pass via options to `RESTClient()`? That way it's more clear what's being initialized, there is no reliance to global state, and we don't have to do `t.Cleanup(gock.Off())`

gh_test.go
```diff
@@ -85,25 +86,26 @@ func TestGQLClient(t *testing.T) {
...
 
-	http := httpmock.NewRegistry(t)
-	http.Register(
-		httpmock.GQL("QUERY"),
-		httpmock.StringResponse(`{"data":{"viewer":{"login":"hubot"}}}`),
-	)
+	gock.New("https://api.github.com").
+		Post("/graphql").
+		MatchHeader("Authorization", "token GH_TOKEN").
+		BodyString("QUERY").
```
mislav: Perhaps we could define a custom matcher that matches the query name in a GraphQL request by parsing the JSON and looking solely at the `query` value, rather than at the whole body of the request

gh_test.go
```diff
@@ -85,25 +86,26 @@ func TestGQLClient(t *testing.T) {
...
 
-	client, err := GQLClient(&api.ClientOptions{Transport: http})
+	client, err := GQLClient(nil)
 	assert.NoError(t, err)
 
 	vars := map[string]interface{}{"var": "test"}
 	res := struct{ Viewer struct{ Login string } }{}
 	err = client.Do("QUERY", vars, &res)
 	assert.NoError(t, err)
+	assert.True(t, gock.IsDone())
```
mislav: If this is false, the resulting test failure won't be very descriptive. Is there a way to output the information about stubs that weren't matched?

mislav (changes requested):
This looks great so far!

gh_test.go
samcoe: I thought we could get around this but [this check](https://github.com/h2non/gock/blob/48c3ba5c871a49c9576f33d5c7d88eff9404d757/transport.go#L48) functionality deep in the `gock` implementation makes it infeasible.

gh_test.go
samcoe: We could do that. What about just matching the whole body here since it is not a long one? Do you feel that the JSON parsing approach provides significant additional value?

gh_test.go
mislav: Ah I see; it looks like the library is designed in a way that it always wants to be used with global intercept. 👍

mislav (approved):
Thanks! Looks great

PR #33, opened 2022-05-06
Add IsAuthenticated

### LINKED ISSUE #32: Check auth status
If the CLI isn't already authenticated, extensions can fail in various ways depending on what they call. For example, using `gh.CurrentRepository()` - even in a repo - fails with `errors.New("unable to determine current repository, none of the git remotes configured for this repository point to a known GitHub host")`. If you specify a repo explicitly, `gh.GQLClient()` can fail with an internal `config.NotFoundError` which, given it's in `internal`, cannot be referenced so might as well be a generic `error`. To the user, the error simply prints "not found" which isn't actionable. Even the more wordy error mentioned first isn't that actionable, though more useful.

Instead, what about something like `gh.IsAuthenticated() bool`? It could, for example, call `config.Load()` and then check the `Config.AuthToken` for a known host. This would allow extensions to easily check up front if they should prompt the user to authenticate.

Alternatively, expose any auth-related errors in `pkg` or something so we can do type assertions.

heaths:
Fixes #32. Also exports error types.

### DIFF
--- a/gh.go
+++ b/gh.go
@@ -7,7 +7,6 @@ package gh
 
 import (
 	"bytes"
-	"errors"
 	"fmt"
 	"net/http"
 	"net/url"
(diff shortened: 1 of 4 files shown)


samcoe:
Closing in favor of #44, please see https://github.com/cli/go-gh/issues/32#issuecomment-1161847775 for context.

PR #35, opened 2022-05-09
Export GraphQL error types

heaths:
Similar to `HTTPError` and `HTTPErrorItem` already (which was actually `HttpErrorItem`), export `GQLError` (was `GQLErrorREsponse`) and `GQLErrorItem` (was `GQLError`). This makes it possible for callers to check the errors and react accordingly, as in cli/cli#5588 where certain GraphQL errors could be filtered out.

### DIFF
--- a/gh_test.go
+++ b/gh_test.go
@@ -105,6 +105,32 @@ func TestGQLClient(t *testing.T) {
 	assert.Equal(t, "hubot", res.Viewer.Login)
 }
 
+func TestGQLClientError(t *testing.T) {
+	tempDir := t.TempDir()
+	orig_GH_CONFIG_DIR := os.Getenv("GH_CONFIG_DIR")
+	orig_GH_TOKEN := os.Getenv("GH_TOKEN")
+	t.Cleanup(func() {
+		os.Setenv("GH_CONFIG_DIR", orig_GH_CONFIG_DIR)
+		os.Setenv("GH_TOKEN", orig_GH_TOKEN)
+	})
+	os.Setenv("GH_CONFIG_DIR", tempDir)
+	os.Setenv("GH_TOKEN", "GH_TOKEN")
+
+	http := httpmock.NewRegistry(t)
+	http.Register(
+		httpmock.GQL("QUERY"),
+		httpmock.StringResponse(`{"errors":[{"type":"NOT_FOUND","path":["organization"],"message":"Could not resolve to an Organization with the login of 'cli'."}]}`),
+	)
+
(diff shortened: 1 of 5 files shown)


heaths:
A specific use case is a work around for https://github.com/github/feedback/discussions/16168. A `ProjectNext` object can exist in either an `Organization` or a `User` if not already linked to a `Repository`. But to minimize the number of requests I would ideally query both the root `organization` and `user` to figure out a `ProjectNext.id`. In fact, given both an owner and repo name, I could query all three root objects and use whichever `ProjectNext` I can, but I need to ignore the other 2 errors.

Technically, the current implementation still populates the response `interface{}` despite returning an `error`, but this may not always be the case since its atypical. Relying solely on the error message is also fragile and again relies on implementation details. Instead, just like `RESTClent`, export the `GQLError` so they can be used by the caller.

pkg/api/client.go
```diff
@@ -97,3 +99,23 @@ type GQLClient interface {
 	// to the GitHub GraphQL schema.
 	Query(name string, query interface{}, variables map[string]interface{}) error
 }
+
+// GQLError contains GQLErrors from a GraphQ request.
```
samcoe:
```suggestion
// GQLError contains GQLErrors from a GraphQL request.
```

samcoe (commented):
This looks good to me. Only comment would be that I think we should rename `HttpErrorItem` -> `HTTPErrorItem` in this PR as well since we are already here and it was a typo to begin with. Since `go-gh` is beta I don't think there is much controversy with fixing that small mistake.

gh_test.go
```diff
@@ -103,6 +103,32 @@ func TestGQLClient(t *testing.T) {
...
+		os.Setenv("GH_TOKEN", orig_GH_TOKEN)
+	})
+	os.Setenv("GH_CONFIG_DIR", tempDir)
+	os.Setenv("GH_TOKEN", "GH_TOKEN")
+
+	http := httpmock.NewRegistry(t)
+	http.Register(
+		httpmock.GQL("QUERY"),
+		httpmock.StringResponse(`{"errors":[{"type":"NOT_FOUND","path":["organization"],"message":"Could not resolve to an Organization with the login of 'cli'."}]}`),
+	)
```
samcoe: I just merged in the migration to `gock` so if we could change this mock to use the new approach that would be great 🙏

heaths:
I did rename it in another commit but realized I forgot to push it. 😣

samcoe (approved):
LGTM!

PR #36, opened 2022-05-09
Add default host as known host when environment auth token exists

### LINKED ISSUE #34: CurrentRepository() fails when executed inside a GH Workflow
Hey - just discovered something that might be interesting.

When a consuming application calls `CurrentRepository()` from inside a GH workflow I'm seeing the following error:

```
unable to determine current repository, none of the git remotes configured for this repository point to a known GitHub host
```

I think this is because `~/.config/gh/hosts.yml` does not exist on a runner because of the TOKEN based auth (vs oauth).

This would cause [FilterByHosts](https://github.com/cli/go-gh/blob/c41a1271d97cf42ba0dcf94289da729cdb1554eb/gh.go#L141)  to return an error.

I have worked around this by. [exporting the GH_HOST](https://github.com/chelnak/gh-changelog/blob/0b8691c1599c8abb047d0f1321c18f8de7b27843/.github/workflows/changelog.yml#L20) environment variable.. however it was not immediately obvious what to do.

If this is expected behaviour we could update the README with a note on compatibility with GH Actions.

Cheers!

samcoe:
This PR adds functionality to `config.Hosts()` to include the default host if there is an environment authentication token that applies to the default host. Additionally it adds tests for `config.Hosts()` that did not exist before.

Fixes https://github.com/cli/go-gh/issues/34

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -72,6 +72,9 @@ func (c config) Hosts() []string {
 	if host := os.Getenv(ghHost); host != "" {
 		hosts.Add(host)
 	}
+	if token, _ := c.AuthToken(defaultHost); token != "" {
+		hosts.Add(defaultHost)
+	}
 	entries := c.hosts.keys()
 	hosts.AddValues(entries)
 	return hosts.ToSlice()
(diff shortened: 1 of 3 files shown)


mislav (approved):
Thanks for fixing this!

PR #38, opened 2022-05-09
Update test and lint workflow files

samcoe:


.github/workflows/lint.yml
```diff
@@ -1,6 +1,29 @@
 name: Lint
 on: [push, pull_request]
+permissions:
```
mislav, line 3: What does "permissions" apply to?

samcoe: I copied from https://github.com/cli/cli/blob/99952e2e271755690b9a003b4c97f4a5548e73c7/.github/workflows/lint.yml#L14, I _think_ this means that any of the jobs in this file do not have write permissions back to the repo.

PR #39, opened 2022-05-11
Add additional API functionality

samcoe:
This PR adds additional API functionality that is necessary for `gh` to utilize `go-gh`. 

* Added `optionsNeedResolution` to avoid unnecessary config loading
* Exported `api.HandleHTTPError` to allow users access to use our error handling
* Added `RESTClient.Raw` function to allow users direct access to an http response
* Renamed `pkg/api/http.go` to `pkg/api/errors.go` to more accurately reflect its contents

Based on top of https://github.com/cli/go-gh/pull/30
Corresponding `gh` PR https://github.com/cli/cli/pull/5614 that uses these changes.


pkg/api/client.go
```diff
@@ -85,6 +83,12 @@ type RESTClient interface {
 	// Put issues a PUT request to the specified path with the specified body.
 	// The response is populated into the response argument.
 	Put(path string, body io.Reader, response interface{}) error
+
+	// Raw issues a request with type specified by method to the
+	// specified path with the specified body.
+	// The response is returned rather than being populated
+	// into a response argument.
+	Raw(method string, path string, body io.Reader) (*http.Response, error)
```
samcoe: I am not in love with this name, I am open to suggestions on a better one.

pkg/api/errors.go
```diff
@@ -0,0 +1,173 @@
...
+
+func matchPath(p, expect string) bool {
+	if strings.HasSuffix(expect, ".") {
+		return strings.HasPrefix(p, expect) || p == strings.TrimSuffix(expect, ".")
+	}
+	return p == expect
+}
+
+// HandleHTTPError parses a http.Response into a HTTPError.
+func HandleHTTPError(resp *http.Response) error {
```
samcoe, line 104: I am not in love with this name, I am open to suggestions on a better one.

pkg/api/client.go
mislav: Ideally it would be something like `Do` to indicate the low-levelness of this command, but that's already taken 😅 I'm open to something generic like "Request". Agreed that "raw" is unclear here

pkg/api/errors.go
mislav: I think it's fine! Ideally, no integrators will ever need to call this directly, since all the other higher-level API methods call this internally on non-HTTP 200 statuses. However, I understand we need this from `cli/cli`?

mislav (approved):
This looks good! Thank you for the hard work 🙇

pkg/api/errors.go
samcoe: Yeah, there are numerous places in `cli/cli` that call `api.HandleHTTPError` directly rather than making requests through an `api.Client` interface.

PR #40, opened 2022-05-16
Export ssh package

samcoe:
This PR exports the SSH package so it can be used inside of `gh`. It also changes the interface a bit now utilize a `Translator` interface as this name is more accurate.

corresponds to https://github.com/cli/cli/pull/5654


mislav (approved):
Looks great!

PR #41, opened 2022-05-16
Add repository.ParseWithHost

samcoe:
This PR adds a `repository.ParseWithHost` function that is necessary for `gh` to use `go-gh` for parsing repository strings. Additionally, it changes `repository.Parse` to now look at the config file for a viable host instead of just looking at `GH_HOST` before defaulting to `github.com`. This new behavior also matches the current functionality in `gh`.

corresponds to https://github.com/cli/cli/pull/5655

### DIFF
--- a/pkg/repository/repository.go
+++ b/pkg/repository/repository.go
@@ -4,9 +4,9 @@ package repository
 
 import (
 	"fmt"
-	"os"
 	"strings"
 
+	"github.com/cli/go-gh/internal/config"
 	"github.com/cli/go-gh/internal/git"
 	irepo "github.com/cli/go-gh/internal/repository"
 )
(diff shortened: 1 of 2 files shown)


pkg/repository/repository.go
```diff
@@ -46,12 +47,48 @@ func Parse(s string) (Repository, error) {
 	case 3:
 		return irepo.New(parts[0], parts[1], parts[2]), nil
 	case 2:
-		host := os.Getenv("GH_HOST")
-		if host == "" {
-			host = "github.com"
+		host := "github.com"
+		cfg, err := config.Load()
```
mislav, line 51: This seems correct for now, but ultimately I find it surprising that `Parse()` implicitly loads config. I think an ideal API might be one where a repository "parser" is derived from a config object, so that their relationship is denoted explicitly.

mislav (approved):
🚀

pkg/repository/repository.go
samcoe: Yeah I agree it is a bit odd. `go-gh` is also targeting extension authors, I would think that they don't necessarily want to concern themselves with loading up the config and determining the correct host. We also do this when initializing the various API clients. As long as there are alternatives for advanced users, that don't want automatic resolution such as `repository.ParseWithHost`, I am okay with having these easy to use alternatives.

PR #44, opened 2022-05-24
Refactor config package and add functionality

samcoe:
This PR introduces a bunch of functionality necessary for `gh` to utilize `go-gh` for config handling.

Things changed:
* Renamed and improved `configmap` to `yamlmap`
* Exported `config` package with functions for manipulating configuration files. Only business logic that is encapsulated is around reading and writing of configuration files.
* Exported `auth` package with functions related to auth tokens and hosts. This functionality is not new, it just used to live in the internal `config` package.

I added inline comments and questions for design decisions.

Note: This PR is not as long as it seems, lots of code was just shuffled around.


pkg/auth/auth.go
```diff
@@ -0,0 +1,80 @@
+package auth
```
samcoe, line 3: Should this be named `auth` or `authentication`?

pkg/config/config.go
```diff
@@ -0,0 +1,269 @@
+// Package config is a set of types for interacting with the gh configuration files.
+// Note: This package is intended for use only in `gh`, any other use cases are subject
+// to breakage and non-backwards compatible updates.
+package config
```
samcoe, line 4: Should this be `config` or `configuration`?

pkg/config/config.go
```diff
@@ -0,0 +1,269 @@
...
+		if err != nil {
+			entry = yamlmap.MapValue()
+			m.AddEntry(key, entry)
+		}
+		m = entry
+	}
+	m.SetEntry(keys[len(keys)-1], yamlmap.StringValue(value))
+}
+
+func Read() (*Config, error) {
```
samcoe, line 108: Went with `Read` instead of `Load` as it matched `Write` better.

pkg/config/config.go
```diff
@@ -0,0 +1,269 @@
...
+	// TODO: Make global config singleton using sync.Once
```
samcoe, line 109: Future improvement here that I didn't wan't to tackle in this PR.

pkg/auth/auth.go
```diff
@@ -0,0 +1,80 @@
...
+	ghEnterpriseToken     = "GH_ENTERPRISE_TOKEN"
+	ghHost                = "GH_HOST"
+	ghToken               = "GH_TOKEN"
+	githubEnterpriseToken = "GITHUB_ENTERPRISE_TOKEN"
+	githubToken           = "GITHUB_TOKEN"
+	oauthToken            = "oauth_token"
+	hostsKey              = "hosts"
+)
+
+func TokenForHost(cfg *config.Config, host string) (string, string) {
```
mislav: Should these top-level functions take `cfg` as argument, or instead just access the `config.Config` instance internally as a singleton?

pkg/auth/auth.go
mislav: `auth` is fine! It reflects the `gh auth` command, which is how end users already manage their lists of hosts and tokens

pkg/config/config.go
```diff
@@ -0,0 +1,269 @@
...
+	file, err := os.OpenFile(filename, os.O_RDWR|os.O_CREATE|os.O_TRUNC, 0600)
+	if err != nil {
+		return err
+	}
+	defer file.Close()
+	_, err = file.Write(data)
+	return err
+}
+
+var defaultGeneralEntries = heredoc.Doc(`
```
mislav: Nit: since there is no leading indentation, could we just use a normal Go string literal without `heredoc.Doc`?

pkg/config/config.go
mislav: I'm fine with `config`!

pkg/config/config.go
```diff
@@ -0,0 +1,269 @@
...
+	xdgConfigHome = "XDG_CONFIG_HOME"
+	xdgDataHome   = "XDG_DATA_HOME"
+	xdgStateHome  = "XDG_STATE_HOME"
+)
+
+type Config struct {
+	entries *yamlmap.Map
+}
+
+func Get(c *Config, keys []string) (string, error) {
```
mislav: Now that `config.Config` is just a struct and not some interface anymore, what are the tradeoffs of having `config.Get/Set/Remove` etc. be methods on the top level instead of on the struct itself?

pkg/config/errors.go
```diff
@@ -0,0 +1,27 @@
+package config
+
+import (
+	"fmt"
+)
+
+// InvalidConfigFileError represents an error when trying to read a config file.
+type InvalidConfigFileError struct {
+	Path string
+	Err  error
```
mislav, line 10: Custom error types that wrap another error should implement `Unwrap()`

pkg/config/errors.go
```diff
@@ -0,0 +1,27 @@
...
+}
+
+// Allow InvalidConfigFileError to satisfy error interface.
+func (e InvalidConfigFileError) Error() string {
+	return fmt.Sprintf("invalid config file %s: %s", e.Path, e.Err)
+}
+
+// NotFoundError represents an error when trying to find a config key
+// that does not exist.
+type NotFoundError struct {
```
mislav: Should this be `KeyNotFoundError`? Otherwise it might read as if it's a "config not found" error, as in a whole config file

mislav (commented):
This looks great! I love the surface simplicity, the separate `yamlmap` package, and the fact that extension authors are discouraged to interact with the core config directly. I have some polish-level comments

pkg/auth/auth.go
samcoe: Yeah I think I agree, these functions feel like they should know how to access the config themselves, I can't think of a scenario where we would want the user to pass in their own `config.Config` instance. My only worry is that `config.Config` becomes an implicit dependency of these functions rather than an explicit one.

pkg/config/config.go
samcoe: That is a good question. I don't know if I can think of a pro/con for either way. Perhaps having them be methods on the struct is more idiomatic? I don't have strong opinions on either direction.

samcoe: After thinking about this more I couldn't come up with a good reason not to make these methods so I made the change.

samcoe:
@mislav This is ready for re-review when you get a chance 🙇

mislav (approved):
This looks great. Thanks for your hard work and for the patience

PR #45, opened 2022-05-30
Turn config into a singleton

samcoe:
This is a follow up PR to https://github.com/cli/go-gh/pull/44. 

Changes in this PR:
- export `ConfigDir`, `StateDir`, `DataDir`
- add Codespaces support in `tokenForHost`
- When using `Read` the loaded config will be cached in a global singleton variable so that subsequent calls to `Read` will not need to access the filesystem again.

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -205,11 +205,11 @@ func load(generalFilePath, hostsFilePath string) (*Config, error) {
 }
 
 func generalConfigFile() string {
-	return filepath.Join(configDir(), "config.yml")
+	return filepath.Join(ConfigDir(), "config.yml")
 }
 
 func hostsConfigFile() string {
-	return filepath.Join(configDir(), "hosts.yml")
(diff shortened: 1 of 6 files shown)


pkg/config/config.go
```diff
@@ -105,10 +120,12 @@ func (c *Config) Set(keys []string, value string) {
 
 // Read gh configuration files from the local file system and
 // return a Config.
-func Read() (*Config, error) {
-	// TODO: Make global config singleton using sync.Once
-	// so as not to read from file every time.
-	return load(generalConfigFile(), hostsConfigFile())
+var Read = func() (*Config, error) {
```
samcoe, line 123: Changing this to be a variable for testing purposes seemed to be the best approach here to bypass having to deal with `sync.Once` in tests.

mislav (approved):
Love how much simpler the test setup is with this 👌

PR #49, opened 2022-06-14
Improvements to API package for use in cli

samcoe:
This PR helps address feedback received on https://github.com/cli/cli/pull/5614.

This PR does four things:
* Adds ability to set api cache options with request headers
* Removes unnecessary preview headers
* Changes `GQL` to `GraphQL` for error messages
* Fixed a test that was failing locally, but not on CI

### DIFF
--- a/internal/api/http.go
+++ b/internal/api/http.go
@@ -168,10 +168,6 @@ func newHeaderRoundTripper(host string, authToken string, headers map[string]str
 		a := "application/vnd.github.merge-info-preview+json"
 		// Preview for visibility when RESTing repos into an org.
 		a += ", application/vnd.github.nebula-preview"
-		// Preview for Commit.statusCheckRollup for old GHES versions.
-		a += ", application/vnd.github.antiope-preview"
-		// Preview for // PullRequest.isDraft for old GHES versions.
-		a += ", application/vnd.github.shadow-cat-preview"
 		headers[accept] = a
 	}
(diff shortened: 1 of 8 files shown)


internal/api/http.go
```diff
@@ -83,16 +83,14 @@ func NewHTTPClient(opts *api.ClientOptions) http.Client {
...
-		}
-		c := cache{dir: opts.CacheDir, ttl: opts.CacheTTL}
-		transport = c.RoundTripper(transport)
+	if opts.CacheDir == "" {
+		opts.CacheDir = filepath.Join(os.TempDir(), "gh-cli-cache")
+	}
+	if opts.EnableCache && opts.CacheTTL == 0 {
+		opts.CacheTTL = time.Hour * 24
 	}
+	c := cache{dir: opts.CacheDir, ttl: opts.CacheTTL}
```
samcoe, line 92: All requests will not hit the cache layer but if `opts.EnableCache` is not set the TTL will be 0 and it is a no-op.

internal/api/http.go
```diff
@@ -83,16 +83,14 @@ func NewHTTPClient(opts *api.ClientOptions) http.Client {
...
+	transport = c.RoundTripper(transport)
```
samcoe, line 93: All requests will now hit the cache layer but if `opts.EnableCache` is not set the TTL will be 0 and it is a no-op.

mislav (approved):
Looks great!

PR #50, opened 2022-06-15
Add context for GraphQL and REST clients

### LINKED ISSUE #47: Add context to possibly long-running methods (external API)
### Describe the feature or problem you’d like to solve

Currently, there is no option to cancel the ongoing request.

Here is a small example of pagination request where this would be helpful https://gist.github.com/mszostok/b68ff95f85d4b4ff8a27aeed56f9d3ca.

This only fetches GitHub stars, so the payload is not heavy, however if you want to fetch all issues then you have even a bigger problem.

### Proposed solution

How will it benefit CLI and its users?

- It improves UX as you can cancel executed command, and it will be released immediately instead of being unresponsive for a few seconds.

- It reduces the quota consumption, as you won't execute next calls if not needed and also save the bandwidth as you will inform server that the client is not waiting for the response anymore.

### Additional context

There are at least two options how it can be achieved:
1. Add new methods with the `WithContext` suffix. For example:

    ```go
    // GQLClient is the interface that wraps methods for the different types of
    // API requests that are supported by the server.
    type GQLClient interface {
      // Do executes a GraphQL query request.
      // The response is populated into the response argument.
      Do(query string, variables map[string]interface{}, response interface{}) error

      // DoWithContext executes a GraphQL query request.
      // The response is populated into the response argument.
      DoWithContext(ctx context.Context, query string, variables map[string]interface{}, response interface{}) error
    }
    ```

    Similar approach as Go has for [http.Request](https://github.com/golang/go/blob/9228d7d7d523e6831933b79f768dfbb51f8ffb5b/src/net/http/request.go#L834-L861). It's not a breaking change.
2. Change all func signature and add the `context.Context` as a first parameter. It's a breaking change in external API.

There is also an option to add [`context.Context` to struct](https://go.dev/blog/context-and-structs) but it won't work in this case and it's ugly and problematic.

If you agree with one of the provided option, I can create a dedicated PR 👍

mszostok:
**Description**

Changes proposed in this pull request:

- Add `context.Context` for GraphQL client. For all methods there is alternative, with pattern `{method_name}WithContext`
  - Add unit test for `context.Canceled` and `context.DeadlineExceeded` to prevent any regression.
- Add `context.Context` for REST client. I added only `DoWithContext` and `RequestWithContext`. Rest methods are just a syntax sugar, and I don't see value into adding an alternative method to all of them. However, it's easy to add, so if you want to, let me know and I will do it 👍
  - Add unit test for `context.Canceled` and `context.DeadlineExceeded` to prevent any regression.

**Related issue(s)**

Fixes #47

### DIFF
--- a/internal/api/gql_client.go
+++ b/internal/api/gql_client.go
@@ -9,6 +9,7 @@ import (
 	"net/http"
 
 	"github.com/cli/go-gh/pkg/api"
+
 	graphql "github.com/cli/shurcooL-graphql"
 )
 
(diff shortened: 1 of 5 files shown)


samcoe (approved):
@mszostok This looks great, thanks for including tests with the PR!

PR #55, opened 2022-07-25
Do not return an error when blank config files are read

samcoe:
This PR changes the behavior when reading empty config files, previously we would return an `ErrInvalidFormat` and now we will return a blank yaml map. This fixes backwards compatibility where some cli users had blank `host.yml` files.

Fixes https://github.com/cli/cli/issues/5949

### DIFF
--- a/pkg/ssh/ssh_test.go
+++ b/pkg/ssh/ssh_test.go
@@ -91,7 +91,6 @@ func Test_sshParser_absolutePath(t *testing.T) {
 		parentFile string
 		arg        string
 		want       string
-		wantErr    bool
 	}{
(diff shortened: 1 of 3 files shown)


pkg/ssh/ssh_test.go
```diff
@@ -91,7 +91,6 @@ func Test_sshParser_absolutePath(t *testing.T) {
 		parentFile string
 		arg        string
 		want       string
-		wantErr    bool
```
samcoe, line 94: This was unused so I removed it, not related to other changes in the PR.

mislav (approved):
Good catch!

PR #56, opened 2022-07-25
Allow skipping option and header resolution for api clients

samcoe:
This PR adds the `SkipResolution` option to `api.ClientOptions` allowing the user to skip all automatic resolutions that `go-gh` will do upon initializing an api client. This PR also changes the "Time-Zone" header to only be added if it was properly resolved, both changes are necessary for fixing [5978](https://github.com/cli/cli/issues/5978).

Fixes https://github.com/cli/cli/issues/5978
cc https://github.com/cli/cli/pull/5935

### DIFF
--- a/internal/api/http.go
+++ b/internal/api/http.go
@@ -112,6 +112,12 @@ func NewHTTPClient(opts *api.ClientOptions) http.Client {
 		transport = logger.RoundTripper(transport)
 	}
 
+	if opts.Headers == nil {
+		opts.Headers = map[string]string{}
+	}
+	if !opts.SkipResolution {
+		resolveHeaders(opts.Headers)
+	}
 	transport = newHeaderRoundTripper(opts.Host, opts.AuthToken, opts.Headers, transport)
 
 	return http.Client{Transport: transport, Timeout: opts.Timeout}
@@ -148,10 +154,7 @@ type headerRoundTripper struct {
 	rt      http.RoundTripper
 }
 
-func newHeaderRoundTripper(host string, authToken string, headers map[string]string, rt http.RoundTripper) http.RoundTripper {
-	if headers == nil {
-		headers = map[string]string{}
-	}
+func resolveHeaders(headers map[string]string) {
 	if _, ok := headers[contentType]; !ok {
 		headers[contentType] = jsonContentType
 	}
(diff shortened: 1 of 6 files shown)


internal/api/http.go
```diff
@@ -167,11 +170,11 @@ func newHeaderRoundTripper(host string, authToken string, headers map[string]str
...
 		}
 	}
-	if _, ok := headers[authorization]; !ok && authToken != "" {
-		headers[authorization] = fmt.Sprintf("token %s", authToken)
-	}
 	if _, ok := headers[timeZone]; !ok {
-		headers[timeZone] = currentTimeZone()
+		tz := currentTimeZone()
+		if tz != "" {
+			headers[timeZone] = tz
```
mislav, line 176: This alone looks like it's sufficient to solve the original issue (sending the blank Time-Zone header). How come additional changes to the functionality are needed?

mislav (commented):
Thanks for fixing, but I wonder whether the fix could be scoped down to a minimal amount of changes. The fact that this adds a new feature to work around the bug feels heavy-handed and potentially risky to me.

internal/api/http.go
samcoe: The reason that is not sufficient to solve the bug is because [this](https://github.com/cli/cli/blob/02881b4783db05be2249c27f36130e3ef82daa9a/api/client.go#L257) line is explicitly telling `go-gh` not to resolve the "Time-Zone" header and leave it set to an empty string. We could detect that in `go-gh` and remove the "Time-Zone" header if it set to an empty string but that felt off since we have been following the rule that any user supplied values are not modified. Adding the `SkipResolution` options feels like a cleaner solution and also addresses the confusion you correctly identified [here](https://github.com/cli/cli/pull/5935#discussion_r921065970).

mislav: Fair enough, but if there is going to be an additional config option, maybe it could be better documented? Right now I don't think it's clear what "resolution" is, since the user probably isn't familiar with the implementation of go-gh. Since the configuration option only affects whether default headers will be added, maybe the option could be called `skipDefaultHeaders`?

mislav (approved):

internal/api/http.go
samcoe: Great suggestion, that is a much more descriptive name.

PR #58, opened 2022-07-25
Correctly determine IANA Time Zone name

mislav:
Maintaining a list of UTC mappings internally wasn't a great approach due to these mappings being inaccurate during DST.

Thanks @natefinch!

Ref. https://github.com/cli/cli/issues/5978
Ref. https://github.com/cli/go-gh/pull/56


samcoe (approved):
I like the idea of not having to maintain the timezone list ourselves 👍

PR #60, opened 2022-08-02
Expand logging support for HTTP traffic

mislav:
The central idea of this change is that GH_DEBUG environment variable is respected the same way it is in gh CLI.

- Respect the GH_DEBUG environment variable by default, but add the option to skip it;
- Add option to opt into colorized logging;
- Add option to explicitly log HTTP headers and bodies—the new default is off;
- Colorize JSON payloads when logging HTTP bodies;
- Log form-encoded payloads;
- Pretty-print GraphQL queries when logging HTTP requests.


pkg/api/client.go
```diff
@@ -38,6 +38,15 @@ type ClientOptions struct {
 	// Default is no logging.
 	Log io.Writer
 
+	// LogIgnoreEnv disables respecting the GH_DEBUG environment variable for logging.
+	LogIgnoreEnv bool
+
+	// LogColorize enables colorized logging to Log for display in a terminal.
```
samcoe, line 45:
```suggestion
	// LogColorize enables colorized logging to Log for display in a terminal.
	// Default is no coloring.
```

With the comments I have been trying to describe the default state for each options as I generally find it useful when I look at documentation.

pkg/api/client.go
```diff
@@ -38,6 +38,15 @@ type ClientOptions struct {
...
+	LogColorize bool
+
+	// LogVerboseHTTP enables logging HTTP headers and bodies.
```
samcoe:
```suggestion
	// LogVerboseHTTP enables logging HTTP headers and bodies.
	// Default is only logging request URLs and response statuses.
```

I am actually unsure what httpretty.Logger does by default but would like to see it described here in the comments.

pkg/jsonpretty/format.go
```diff
@@ -0,0 +1,143 @@
+package jsonpretty
```
samcoe, line 2: Do we want this package to be exported? If so, can we add a comment to the top here describing the purpose and functionality of the package?

pkg/api/client.go
```diff
@@ -38,6 +38,15 @@ type ClientOptions struct {
 	// Default is no logging.
 	Log io.Writer
 
+	// LogIgnoreEnv disables respecting the GH_DEBUG environment variable for logging.
+	LogIgnoreEnv bool
```
samcoe, line 43: After thinking about this more, I am not sure that this option is necessary. I understand its purpose, but can't think of a good use case where we want to allow an extension author to ignore `GH_DEBUG`. 

Additionally, I think our users would find the behavior of this option a bit confusing without further comments describing its behavior.

samcoe (changes requested):
Overall the code looks really good to me. I left a couple questions/comments mostly around documentation.

pkg/api/client.go
mislav: > After thinking about this more, I am not sure that this option is necessary. I understand its purpose, but can't think of a good use case where we want to allow an extension author to ignore `GH_DEBUG`.

Thanks for chiming in. One consumer of go-gh that wants to use this option is the CLI itself, since GitHub CLI already parses environment variables itself and wants to prevent underlying libraries to parse the same environment variables, lest some inconsistencies or conflicts arise.

Additionally, an extension author might want to disable env-controlled logging if it already offers the user granular controls for logging.

As a general rule of thumb, if a generic Go library automatically respects environment variables, it's a good idea to offer a way for the consumer of said library to skip respecting environment variables in cases where that is not desired. At the end of the day, environment variables are essentially global variables shared across modules and even processes, and to mitigate their potential harms, developers should have ways to put up boundaries to limit the propagation (or effect) of these variables.

I have added some extra documentation detailing this configuration option.

samcoe (approved):
Thanks for addressing the requested changes 🙇

pkg/api/client.go
samcoe: That all makes sense to me, my only thought was that if an extension author wanted to ignore `GH_DEBUG` they could provide a value for the `Log` option making `LogIgnoreEnv` feel a bit redundant.

PR #61, opened 2022-08-02
Extract table printer from CLI

mislav:
This exposes a TablePrinter that can be used to print column-formatted tabular data to the terminal and the same data in TSV format to redirect to files or scripts.

TODO
- [x] more tests
- [x] API docs

### DIFF
--- a/example_gh_test.go
+++ b/example_gh_test.go
@@ -8,6 +8,8 @@ import (
 	"time"
 
 	"github.com/cli/go-gh/pkg/api"
+	"github.com/cli/go-gh/pkg/tableprinter"
+	"github.com/cli/go-gh/pkg/term"
 	graphql "github.com/cli/shurcooL-graphql"
 )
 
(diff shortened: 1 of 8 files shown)


example_gh_test.go
```diff
@@ -162,3 +164,25 @@ func ExampleCurrentRepository() {
...
+
+	// add a field that will render with color and will not be auto-truncated
+	t.AddField("1", tableprinter.WithColor(red), tableprinter.WithTruncate(nil))
+	t.AddField("hello")
+	t.EndRow()
+	t.AddField("2")
+	t.AddField("world")
+	if err := t.Render(); err != nil {
+		fmt.Fprintln(os.Stderr, err)
+		os.Exit(1)
```
samcoe: I think this is equivalent to `log.Fatal(err)`.

pkg/tableprinter/table.go
```diff
@@ -0,0 +1,238 @@
...
+
+type fieldOption func(*tableField)
+
+type TablePrinter interface {
+	AddField(string, ...fieldOption)
+	EndRow()
+	Render() error
+}
+
+// WithTruncate sets the truncation function for the field. Pass nil to disable truncation for this value.
```
samcoe: If no `WithTruncate` function is specified for a field will the the field be truncated as necessary to fit the output width?

pkg/term/env.go
```diff
@@ -0,0 +1,144 @@
...
+}
+
+// IsTerminalOutput returns true if standard output is connected to a terminal.
+func (t Term) IsTerminalOutput() bool {
+	return t.isTTY
+}
+
+// ColorEnabled reports whether it's safe to output ANSI color sequences, depending on IsTerminalOutput and
+// environment variables.
+func (t Term) ColorEnabled() bool {
```
samcoe: Do we want to name this `IsColorEnabled` to match naming scheme of `IsTerminalOutput`?

samcoe (approved):
This looks great! Added just a couple non-blocking comments/questions.

pkg/tableprinter/table.go
mislav: That's a good question! Every field is automatically truncated by default. I will try to point this out in the docs

pkg/term/env.go
mislav: Good idea!

PR #62, opened 2022-08-08
Migrate browser package from gh

samcoe:
This PR extracts out the `Browser` package from `gh`. Functionality wise it remains the same as it was in `gh`. After this is merged I will make a follow up PR in [cli/cli](https://github.com/cli/cli) to use this extracted package. 

### DIFF
--- a/pkg/browser/browser.go
+++ b/pkg/browser/browser.go
@@ -1,0 +1,80 @@
+// Package browser facilitates opening of URLs in a web browser.
+package browser
+
+import (
+	"io"
+	"os"
+	"os/exec"
+
+	cliBrowser "github.com/cli/browser"
+	"github.com/cli/go-gh/pkg/config"
+	"github.com/cli/safeexec"
+	"github.com/google/shlex"
+)
+
+// Browser represents a web browser that can be used to open up URLs.
+type Browser struct {
+	launcher string
+	stderr   io.Writer
+	stdout   io.Writer
+}
+
+// NewBrowser initializes a Browser. If a launcher is not specified
+// one is determined based on environment variables or from the
+// configuration file.
+// The order of precedence for determining a launcher is:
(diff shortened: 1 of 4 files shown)


pkg/browser/browser.go
```diff
@@ -0,0 +1,80 @@
...
+
+// NewBrowser initializes a Browser. If a launcher is not specified
+// one is determined based on environment variables or from the
+// configuration file.
+// The order of precedence for determining a launcher is:
+// - Specified launcher;
+// - GH_BROWSER environment variable;
+// - browser option from configuration file;
+// - BROWSER environment variable.
+func NewBrowser(launcher string, stdout, stderr io.Writer) Browser {
```
mislav: Since we are in the `browser` package, is it redundant to call the initializer function `NewBrowser`, or would it suffice to just have `New`? I'm fine either way; I'm just wondering about our stance on naming across micro-packages.

mislav (approved):
Looks great 🚀

pkg/browser/browser.go
```diff
@@ -0,0 +1,80 @@
...
+	return b
+}
+
+// Browse opens the launcher and navigates to the specified URL.
+func (b *Browser) Browse(url string) error {
+	return b.browse(url, nil)
+}
+
+func (b *Browser) browse(url string, env []string) error {
+	if b.launcher != "" {
```
vilmibm: nitpick but would be nice to invert this:

```go
if b.launcher == "" {
  return cliBrowser.OpenURL(url)
}
// ... the rest
```

vilmibm (approved):
just a little style comment

pkg/browser/browser.go
samcoe: Agreed, good call. I like `New` better.

PR #63, opened 2022-08-09
Use Go 1.18 for linting

mislav:


samcoe (commented):
What do you think about completely moving to 1.18, since 1.16 which we are currently using in the mod file is deprecated?

mislav:
> What do you think about completely moving to 1.18

Agreed! Note that moving to 1.18 would allow us to unknowingly introduce some Go 1.18-only changes, which would be backwards-incompatible for anyone who is currently building their gh extension with 1.16. If we were post-1.0 already, we would have to be careful with such a change (and possibly bump the major version number), but since we're still in beta, I think we could upgrade the minimum required Go version without heavy repercussions.

PR #67, opened 2022-08-16
Extract export package from gh

samcoe:
This PR extracts out the `export` package from `gh`. Functionality wise, there are minimal changes compared to the counterpart packages in `gh`. After this is merged I will make a follow up PR in [cli/cli](https://github.com/cli/cli) to use this extracted package. 


pkg/template/template.go
```diff
@@ -0,0 +1,232 @@
...
+func New(w io.Writer, width int, colorEnabled bool) Template {
+	return Template{
+		colorEnabled: colorEnabled,
+		output:       w,
+		tp:           tableprinter.New(w, true, width),
+		width:        width,
+	}
+}
+
+func (t *Template) Parse(tpl string) error {
```
samcoe: This is the only change from the `cli` version of this package. `Parse` now needs to be explicitly called rather than implicitly being called from within `Execute`. I made this change to reduce the number of arguments to `New` and `Execute`. Additionally, I find the explicitness of it to be easier to follow.

pkg/jq/jq.go
```diff
@@ -0,0 +1,93 @@
...
+	"io"
+	"math"
+	"os"
+	"strconv"
+
+	"github.com/itchyny/gojq"
+)
+
+// Evaluate a jq expression against an input and write it to an output.
+func Evaluate(input io.Reader, output io.Writer, expr string) error {
```
samcoe, line 16: Open to naming suggestions here? I also changed the argument order to have the `input` come first compared to the same function in `gh`.

pkg/template/template.go
```diff
@@ -0,0 +1,232 @@
...
+
+	var data interface{}
+	if err := json.Unmarshal(jsonData, &data); err != nil {
+		return err
+	}
+
+	return t.tmpl.Execute(t.output, data)
+}
+
+func (t *Template) Flush() error {
```
samcoe, line 94: This used to be called `End` but I felt `Flush` was a more clear representation of what it does.

pkg/jq/jq.go
vilmibm: naming makes sense to me

vilmibm (approved):
looks great, thanks for being clear about what changed.

PR #69, opened 2022-08-25
Extract text package from gh

samcoe:
This PR extracts the text package from `gh`. It also includes `FuzzyAgo` and `FuzzyAgoAbbr` from the `utils` package because they seemed to fit in well here.


pkg/template/template.go
```diff
@@ -207,43 +206,16 @@ func timeAgo(ago time.Duration) string {
 		return "just now"
```
samcoe, line 210: Kind of weird that we don't use `FuzzyAgo` instead of this entire function. Perhaps we should be? The current functionality was brought over from `gh`.

pkg/text/text.go
```diff
@@ -0,0 +1,178 @@
...
+	ellipsis            = "..."
+	minWidthForEllipsis = len(ellipsis) + 2
+)
+
+var indentRE = regexp.MustCompile(`(?m)^`)
+var whitespaceRE = regexp.MustCompile(`\s+`)
+
+// Indent returns a copy of the string s with indent prefixed to it, will indent each line
+// in the string.
+func Indent(s, indent string) string {
```
samcoe, line 23: Open to better naming here. Additionally should we remove any currently prefixed whitespace from each line before adding the indent? That would make it so there is consistent indenting for each line of the string. This is not always what users would want, but perhaps it is the most common use case?

pkg/text/text.go
```diff
@@ -0,0 +1,178 @@
...
+	return fmt.Sprintf("%d %ss", num, thing)
+}
+
+func fmtDuration(amount int, unit string) string {
+	return fmt.Sprintf("about %s ago", Pluralize(amount, unit))
+}
+
+// FuzzyAgo returns a human readable string of the time duration between a and b that is estimated
+// to the nearest unit of time.
+func FuzzyAgo(a, b time.Time) string {
```
samcoe: I would be open to new naming here.

pkg/text/text.go
```diff
@@ -0,0 +1,178 @@
...
+	if ago < 365*24*time.Hour {
+		return fmtDuration(int(ago.Hours())/24/30, "month")
+	}
+
+	return fmtDuration(int(ago.Hours()/24/365), "year")
+}
+
+// FuzzyAgoAbbr is an abbreviated version of FuzzyAgo. It returns a human readable string of the
+// time duration between a and b that is estimated to the nearest unit of time.
+func FuzzyAgoAbbr(a, b time.Time) string {
```
samcoe: I would be open to new naming here.

pkg/text/text_test.go
```diff
@@ -0,0 +1,487 @@
...
+	}
+	for createdAt, expected := range cases {
+		d, err := time.Parse(form, createdAt)
+		assert.NoError(t, err)
+		fuzzy := FuzzyAgoAbbr(now, d)
+		assert.Equal(t, expected, fuzzy)
+	}
+}
+
+func ExampleTruncate() {
```
samcoe: TODO: Fill this in.

samcoe (commented):
I feel pretty good about this extraction, only thing I am a bit concerned about is the clarity of the comments. Any feedback there is appreciated.

AmEarth (commented):
text-package

pkg/text/text.go
mislav: I wouldn't change anything about the input text except apply the indent to each line. Some of the lines of input may already be indented, and as a result they should have double indent to keep everything in the result aligned.

pkg/text/text.go
```diff
@@ -0,0 +1,178 @@
...
+// in the string.
+func Indent(s, indent string) string {
+	if len(strings.TrimSpace(s)) == 0 {
+		return s
+	}
+	return indentRE.ReplaceAllLiteralString(s, indent)
+}
+
+// CamelToKebab returns a copy of the string s that is converted from camel case form to '-' separated form.
+func CamelToKebab(s string) string {
```
mislav: I don't feel that this will be particularly useful for extensions. In fact, it's only ever used in a single place in the CLI codebase, so I guess we could just inline it there.

pkg/text/text.go
```diff
@@ -0,0 +1,178 @@
...
+	}
+	if len(inrune) != 0 {
+		inrune = append(inrune, '-')
+	}
+	inrune = append(inrune, segment...)
+	return inrune
+}
+
+// Title returns a copy of the string s with all Unicode letters that begin words mapped to their Unicode title case.
+func Title(s string) string {
```
mislav: I don't think this would be very useful, since it's trivial to do by hand with `golang.org/x/text/cases` if needed and we don't provide any value on top of it

pkg/text/text.go
```diff
@@ -0,0 +1,178 @@
...
+	c := cases.Title(language.English)
+	return c.String(s)
+}
+
+// RemoveExcessiveWhitespace returns a copy of the string s with excessive whitespace removed.
+func RemoveExcessiveWhitespace(s string) string {
```
mislav: This function is mainly useful when generating table display, and seeing how we already have a package to generate table output with, I don't think we should expose this

pkg/text/text.go
```diff
@@ -0,0 +1,178 @@
...
+	if DisplayWidth(r) < maxWidth {
+		r += " "
+	}
+	return r
+}
+
+// TruncateMultiline returns a copy of the string s that has been shortened to fit the maximum
+// display width. If string s has multiple lines the first line will be shortened and all others
+// removed.
+func TruncateMultiline(maxWidth int, s string) string {
```
mislav: This is specific to our own needs, so I wouldn't expose this for simplicity and just have extension developers rely on Truncate instead

pkg/text/text.go
```diff
@@ -0,0 +1,178 @@
...
+	if i := strings.IndexAny(s, "\r\n"); i >= 0 {
+		s = s[:i] + ellipsis
+	}
+	return Truncate(maxWidth, s)
+}
+
+// Pluralize returns a concatenated string with num and the plural form of thing if necessary.
+func Pluralize(num int, thing string) string {
```
mislav, line 53: I feel that such simplistic pluralization implementation should best be inlined in people's extensions (if needed at all) instead of being exposed from go-gh

pkg/text/text.go
mislav: RelativeTimeAgo

pkg/text/text.go
mislav: I feel like we should have only a single implementation of FuzzyAgo, to reduce confusion. Since the output of this function is very terse, we could drop it in favor of the first implementation

pkg/text/text.go
```diff
@@ -0,0 +1,178 @@
...
+	}
+	if ago < 30*24*time.Hour {
+		return fmt.Sprintf("%d%s", int(ago.Hours())/24, "d")
+	}
+
+	return b.Format("Jan _2, 2006")
+}
+
+// Humanize returns a copy of the string s that replaces all instance of '-' and '_' with spaces.
+func Humanize(s string) string {
```
mislav: Like with CamelToKebab, I don't think this Rails-style helper is useful for gh extension authors

pkg/text/text.go
```diff
@@ -0,0 +1,178 @@
...
+			return ' '
+		}
+		return r
+	}
+	return strings.Map(h, s)
+}
+
+// DisplayURL returns a copy of the string urlStr removing everything except the hostname and path.
+// If there is an error parsing urlStr then urlStr is returned without modification.
+func DisplayURL(urlStr string) string {
```
mislav: This is specific to the needs of the GitHub CLI codebase, but I don't think it should be exposed here. Each extension author could decide how they want to display URLs. If we want to control the experience of how browser URLs are displayed in extensions, we could extend our `browser` package and add the optional feature there.

mislav (changes requested):
I would vote to remove most of the functions there and focus on just the handful of ones that we want to control the implementation of, such as ANSI-codes aware implementations of measuring display width, truncation, and indentation.

Unlike in Ruby or JavaScript, offering a grab-bag of tiny utilities as a Go library isn't what most Go developers want. In fact, if they need some of these utilities in their own programs, most Go developers might choose to just inline them there instead of adding a dependency. Of course, gh extension authors will already depend on the go-gh repository anyway, but at least we can reduce the surface of the APIs we offer.

pkg/text/text.go
samcoe: I agree that it is simplistic, but it is necessary for `FuzzyAgo` and also used independently of that in `gh` so if it isn't exposed here then it will need to be duplicated in `gh`. Going to leave it in for now.

samcoe (commented):
@mislav Thanks for the review. I made the requested changes if you wouldn't mind giving it another look :bow:

mislav (approved):
Thank you! 🚀

PR #71, opened 2022-09-13
Add govulncheck action

samcoe:
This PR adds an action for the new [`govulncheck` tool](https://pkg.go.dev/golang.org/x/vuln/cmd/govulncheck) to check for vulnerabilities in our dependencies.


mislav (commented):
TIL `govulncheck`! How does it compare to Dependabot security alerts, which we currently have enabled?

samcoe:
@mislav So there are two major difference from my understanding:

1. `govulncheck` analyzes your codebase and only surfaces vulnerabilities that actually affect you, based on which functions in your code are transitively calling vulnerable functions.
2. It uses a different database of vulnerabilities to check against. I would guess that there is huge overlap between it and the dependabot one.

I mostly want to enable this side by side with `dependabot` to see if one catches vulnerabilities that the other does not. Long term I can see us deciding to use just one or the other.

.github/workflows/vuln.yml
```diff
@@ -0,0 +1,28 @@
+name: Vulnerability Check
```
mislav: Since this has exactly the same triggers as the `Code Scanning` workflow, I would say that this could be folded in as a separate job under that workflow, rather than as a workflow of its own. At the end of the day, they both perform static analysis to find vulnerabilities

.github/workflows/lint.yml
```diff
@@ -12,7 +10,7 @@ jobs:
       - name: Set up Go
         uses: actions/setup-go@v3
         with:
-          go-version: "1.18"
+          go-version: "1.19"
```
mislav, line 13: Since we still run our tests on 1.18 maybe it would make sense to continue running linter an govulncheck on 1.18 as well. Or do you think matching these versions doesn't matter?

mislav (approved):

.github/workflows/lint.yml
samcoe: I don't think it makes much difference. I purposefully put the linter to 1.19 as there are new godoc features/rules in 1.19 that I wanted to start enforcing in this repo.

PR #72, opened 2022-09-14
Add hyperlink template function

heaths:
Relates to cli/cli#6179

### DIFF
--- a/pkg/template/template.go
+++ b/pkg/template/template.go
@@ -25,6 +25,7 @@ const (
 // Template is the representation of a template.
 type Template struct {
 	colorEnabled bool
+	isTTY        bool
 	output       io.Writer
 	tmpl         *template.Template
 	tp           tableprinter.TablePrinter
@@ -32,9 +33,10 @@ type Template struct {
 }
 
 // New initializes a Template.
-func New(w io.Writer, width int, colorEnabled bool) Template {
+func New(w io.Writer, width int, colorEnabled, isTTY bool) Template {
 	return Template{
 		colorEnabled: colorEnabled,
+		isTTY:        isTTY,
 		output:       w,
 		tp:           tableprinter.New(w, true, width),
 		width:        width,
@@ -47,6 +49,7 @@ func (t *Template) Parse(tmpl string) error {
 	templateFuncs := map[string]interface{}{
 		"autocolor": colorFunc,
 		"color":     colorFunc,
+		"hyperlink": hyperlinkFunc(t.isTTY),
 		"join":      joinFunc,
 		"pluck":     pluckFunc,
 		"tablerender": func() (string, error) {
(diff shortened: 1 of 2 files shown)


pkg/template/template.go
```diff
@@ -25,16 +25,18 @@ const (
...
+	isTTY        bool
 	output       io.Writer
 	tmpl         *template.Template
 	tp           tableprinter.TablePrinter
 	width        int
 }
 
 // New initializes a Template.
-func New(w io.Writer, width int, colorEnabled bool) Template {
+func New(w io.Writer, width int, colorEnabled, isTTY bool) Template {
```
mislav: I'm not sure if we should add an isTTY property to a template. So far, we've kept template output static, no matter where it is connected to. All template helpers have predictable outputs, and the only one that changes dynamically is `autocolor`.

I'd say that `hyperlink` should always output the hyperlink.

pkg/template/template.go
```diff
@@ -233,3 +236,19 @@ func truncateMultiline(maxWidth int, s string) string {
...
+			return link
+		}
+	}
+
+	return func(link, text string) string {
+		if text == "" {
+			text = link
+		}
+
+		return fmt.Sprintf("\x1b]8;;%s\x1b\\%s\x1b]8;;\x1b\\", link, text)
```
mislav: Could this link to the hyperlink "spec"? Thanks

mislav (changes requested):

pkg/template/template.go
heaths: Fair, but then aren't we pushing the work to everyone who uses a template? Alternatively, what if we add an `isTTY` (or something) function that template conditions can use? Or is the assumption that the caller is always in control of both the template and the output, so they can easily and always adapt as needed?

heaths: I remembered that `tableprinter.New` takes an `isTTY` and did respond appropriately so that it was TSV otherwise. Feels like this was regressed. Perhaps we should take an `isTTY`? Or was it ever further discussed to move much/all of `iostreams` to go-gh? @samcoe and I had briefly mentioned it in an issue - I don't think it was even this repo.

Alternatively, if the concern is breaking changes that could grow in time, what about using the `WithOptions` pattern that `tableprinter` is using as well?

heaths:
@mislav I made the changes you requested, but I think it's least worth considering (perhaps not for this PR) whether more should be done in the template functions like they were in cli/cli. Now every script - every invocation - has to vary the template based on *expected* output. Maybe that actually ends up being rare enough since the target of `gh` is probably often known (you're either scripting with it for data, or displaying results directly as part of a script or command line invocation).

samcoe (approved):

mislav (approved):
Awesome, thank you!

pkg/template/template.go
mislav: I think that last time we've had this discussion was in https://github.com/cli/cli/pull/3519#issuecomment-900592822

We've opted to always default to TTY output for consistency so that the user of `--template` doesn't have to guess the actual format of the final output depending on the end-user's execution context. If the gh developer wants separate formats for TTY vs script output, yes in theory they should use two separate templates. But I think `--template` should be mostly for human-readable outputs for terminals anyway.

PR #74, opened 2022-09-20
Allow defining template user functions

### LINKED ISSUE #73: Template support for additional functions
It'd be great if we could add more functions to the `template` package, perhaps using the "WithOptions" pattern as mentioned [elsewhere](https://github.com/cli/go-gh/pull/72#discussion_r972409061). The existing templates are useful, but for extensions to add other useful helpers for their own needs, passing in a map to merge after initializing the default list (so extensions could override) could prove useful.

heaths:
Resolves #73

### DIFF
--- a/pkg/template/template.go
+++ b/pkg/template/template.go
@@ -29,6 +29,7 @@ type Template struct {
 	tmpl         *template.Template
 	tp           tableprinter.TablePrinter
 	width        int
+	funcs        template.FuncMap
 }
 
 // New initializes a Template.
@@ -38,7 +39,14 @@ func New(w io.Writer, width int, colorEnabled bool) Template {
 		output:       w,
 		tp:           tableprinter.New(w, true, width),
 		width:        width,
+		funcs:        template.FuncMap{},
 	}
+}
+
+// RegisterFunc registers a named function or overwrites a built-in function.
+// Call this before Parse.
+func (t *Template) RegisterFunc(name string, f func(fields ...interface{}) (string, error)) {
+	t.funcs[name] = f
 }
 
 // Parse the given template string for use with Execute.
@@ -69,6 +77,9 @@ func (t *Template) Parse(tmpl string) error {
 	}
 	if !t.colorEnabled {
 		templateFuncs["autocolor"] = autoColorFunc
+	}
+	for name, f := range t.funcs {
+		templateFuncs[name] = f
 	}
 	var err error
 	t.tmpl, err = template.New("").Funcs(templateFuncs).Parse(tmpl)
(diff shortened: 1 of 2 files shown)


pkg/template/template.go
```diff
@@ -38,9 +39,16 @@ func New(w io.Writer, width int, colorEnabled bool) Template {
 		output:       w,
 		tp:           tableprinter.New(w, true, width),
 		width:        width,
+		funcs:        template.FuncMap{},
 	}
 }
 
+// RegisterFunc registers a named function or overwrites a built-in function.
+// Call this before Parse.
+func (t *Template) RegisterFunc(name string, f func(fields ...interface{}) (string, error)) {
```
samcoe: Only question I have is should we change this to be `Funcs` and have the argument be a `template.FuncMap` to match the `text/template` package? My vote would be yes to keep consistency with the `text/template` package since we have been doing that with `Parse` and `Execute`. I know I was the one who suggested `RegisterFunc` in the first place 😞

heaths: I've been trying to think of something along these lines as well. Having written an example for testing, the user ends up having to do a lot of what `text/template` does internally.

Do we support devs calling `RegisterFuncs` multiple times and just merge the maps? Last wins, which would be consistent with possibly overriding defaults?

samcoe: That is what I was thinking. We support calling `Funcs` multiple times and just merging the maps where the latest entry for a key wins. That is how `Funcs` works in `text/template`.

samcoe: @heaths I went ahead and made the changes I suggested. Let me know what you think.

samcoe (approved):

pkg/template/template.go
```diff
@@ -38,9 +39,21 @@ func New(w io.Writer, width int, colorEnabled bool) Template {
...
 		width:        width,
+		funcs:        template.FuncMap{},
 	}
 }
 
+// Funcs adds the elements of the argument map to the template's function map.
+// It must be called before the template is parsed.
+// It is legal to overwrite elements of the map including default functions.
+// The return value is the template, so calls can be chained.
+func (t *Template) Funcs(funcMap map[string]interface{}) *Template {
```
heaths, line 50: Nit: I liked `RegisterFuncs` since it was a verb and more indicative of what it does.

heaths (commented):
Thanks for making the changes. It was on my TODO list but a bit down the priority list.

pkg/template/template.go
samcoe: Yeah I went back and forth on that as well. I decided to just copy `text/template` method names as other gophers might be more familiar with it.

PR #81, opened 2022-10-10
Fix non-Windows terminals assumed to always support 256-color

mislav:
Backports https://github.com/cli/cli/pull/6356

### DIFF
--- a/pkg/term/console.go
+++ b/pkg/term/console.go
@@ -4,11 +4,12 @@
 package term
 
 import (
+	"errors"
 	"os"
 )
 
 func enableVirtualTerminalProcessing(f *os.File) error {
-	return nil
+	return errors.New("not implemented")
 }
 
 func openTTY() (*os.File, error) {
--- a/pkg/term/env_test.go
+++ b/pkg/term/env_test.go
@@ -1,0 +1,114 @@
+// Package term provides information about the terminal that the current process is connected to (if any),
+// for example measuring the dimensions of the terminal and inspecting whether it's safe to output color.
+package term
+
+import (
+	"testing"
+)
+
+func TestFromEnv(t *testing.T) {
+	tests := []struct {
+		name          string
+		env           map[string]string
+		wantTerminal  bool
+		wantColor     bool
+		want256Color  bool
+		wantTrueColor bool
+	}{
+		{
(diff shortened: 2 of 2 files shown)


vilmibm (approved):

samcoe (approved):
Code looks good to me. Interesting that govulncheck found an vulnerability that dependabot didn't. The vulnerability wasn't introduced in this PR so it is probably okay to merge, and we can fix the vulnerability in a follow up PR.

heaths:
Wouldn't this completely break alternate screen buffers on *nix then? There are other ways of detecting whether VT sequences are supported:

* Using `tput` e.g., https://unix.stackexchange.com/questions/450365/check-if-terminal-supports-24-bit-true-color
* https://github.com/microsoft/terminal/issues/1040#issuecomment-1043650915

See https://github.com/cli/cli/pull/6356#issuecomment-1276773296 for more information.

heaths:
I submitted https://github.com/cli/cli/pull/6435 to restore behavior without impacting this PR, since the change was only to restore the original behavior without assumptions of VT support this change inferred. This PR would actually be a good fix, but long-term support for various color or other VT support might be good to add to go-gh. See https://github.com/cli/cli/pull/6435#issuecomment-1277776665 for possibilities.

mislav:
@heaths I don't see how this can break alternate screen buffers on *nix. There is still an issue with that in the CLI repo, I agree, but we will resolve that in that repo.

heaths:
Agreed. Please see my latest comment.

heaths:
To clarify my comment about VT support, I was suggesting maybe something in this module could define, but perhaps that would best be in one of @mattn's many term modules, if it isn't already (I couldn't find one).

mattn:
Can I help you? If you can explain how to reproduce the problem, I can fix that.

heaths:
Mainly, with all the terminal modules you have, was wondering if any of them might be a good place to put better VT support detection into. See https://github.com/cli/cli/pull/6435#issuecomment-1277776665 for some thoughts. We've also been wrestling with some of these ideas for Windows Terminal as well. If you think one of your modules might be appropriate, maybe you or I could open an issue and discuss more there.

PR #82, opened 2022-10-13
Fix HTTP error handling in `RESTClient.Request()` method

mislav:
- Ensure that `Request()` calls HandleHTTPError so that any JSON error body properly gets parsed;
- Ensure that response body is closed after reading the error response.

Ref. https://github.com/cli/cli/issues/6429

### DIFF
--- a/internal/api/rest_client.go
+++ b/internal/api/rest_client.go
@@ -33,16 +33,13 @@ func (c restClient) RequestWithContext(ctx context.Context, method string, path 
 
 	resp, err := c.client.Do(req)
 	if err != nil {
-		return resp, err
+		return nil, err
 	}
 
 	success := resp.StatusCode >= 200 && resp.StatusCode < 300
 	if !success {
-		err = api.HTTPError{
-			Headers:    resp.Header,
-			RequestURL: resp.Request.URL,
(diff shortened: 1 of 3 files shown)


internal/api/rest_client.go
```diff
@@ -63,16 +60,17 @@ func (c restClient) DoWithContext(ctx context.Context, method string, path strin
...
 	success := resp.StatusCode >= 200 && resp.StatusCode < 300
 	if !success {
+		defer resp.Body.Close()
 		return api.HandleHTTPError(resp)
 	}
 
 	if resp.StatusCode == http.StatusNoContent {
 		return nil
 	}
+	defer resp.Body.Close()
```
samcoe, line 73: If we move lines 70-72 above line 64 we can remove one of the `defer resp.Body.Close()` lines.

samcoe (approved):
LGTM, just left one comment about some code simplifying. Thanks for cleaning this up.

internal/api/rest_client.go
mislav: In case of HTTP 204 No Content, `resp.Body` will be nil. This is why I'm closing the body in two different places— it's being consumed in two different places too

PR #84, opened 2022-10-21
Resolve ssh hostname aliases with `ssh -G`

mislav:
Going through the `ssh` executable ensures that hostnames get resolved using ssh's exact mechanism and alleviates the need for manually parsing and interpreting ssh configuration files.

I learned this trick from @cmbrose in https://github.com/cli/cli/pull/5958.

Ref. https://github.com/cli/cli/issues/6455

### DIFF
--- a/pkg/ssh/ssh.go
+++ b/pkg/ssh/ssh.go
@@ -1,79 +1,37 @@
-// Package ssh is a set of types and functions for parsing and
-// applying a user's SSH hostname aliases.
+// Package ssh resolves local SSH hostname aliases.
 package ssh
 
 import (
 	"bufio"
-	"io"
 	"net/url"
-	"os"
-	"path/filepath"
-	"regexp"
+	"os/exec"
 	"strings"
+	"sync"
+
+	"github.com/cli/safeexec"
 )
 
-var (
-	configLineRE = regexp.MustCompile(`\A\s*(?P<keyword>[A-Za-z][A-Za-z0-9]*)(?:\s+|\s*=\s*)(?P<argument>.+)`)
-	tokenRE      = regexp.MustCompile(`%[%h]`)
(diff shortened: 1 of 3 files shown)


pkg/ssh/ssh.go
```diff
@@ -84,101 +42,53 @@ func (c config) Translate(u *url.URL) *url.URL {
...
-		file, err = p.open(fileName)
-		if err != nil {
-			return err
-		}
+func (t *Translator) resolve(hostname string) (string, error) {
+	t.cacheMu.RLock()
+	cached, cacheFound := t.cacheMap[strings.ToLower(hostname)]
+	t.cacheMu.RUnlock()
+	if cacheFound {
+		return cached, nil
```
samcoe, line 52: As far as I can tell this cache is never actually populated in this method. I am wondering about its purpose though. We can rectify this and start populating the cache or I think it would be okay to just allow the users of this package to implement the caching since our implementation is just an in-memory map. What are your thoughts?

samcoe (changes requested):
Looks mostly good. This way of resolving SSH hostname aliases is much better than before IMO. I just left one comment regarding the caching mechanism that I would like to see addressed.

pkg/ssh/ssh.go
mislav: Good catch. If the cache is not populated, that was an oversight on my part.

I do think that an in-memory cache would be beneficial since `Translate()` is typically used to translate many URLs at once: think of a git repo that has multiple SSH remotes—Translate will be called 2 times per each remote. Since most remotes will be to the same hostname, a built-in cache avoids shelling out to `ssh` repeatedly.

samcoe (approved):
Thanks for making the requested changes!

PR #85, opened 2022-10-21
Add support for `garage.github.com`

mislav:


samcoe (approved):
Code looks good to me. Kind of odd that the check for `garage` comes before the call to `normalizeHostname()`. Not sure if that can/should be rectified. 

Additionally, would like to see a test to two for this new behavior.

mislav:
> Kind of odd that the check for `garage` comes before the call to `normalizeHostname()`.

Normalizing the hostname turns `garage.github.com` into `github.com`, so it's a bit too late to check for garage afterwards. However, normalizing `garage.github.com` as `github.com` elsewhere is beneficial for token reuse: since it's technically not possible to complete the web browser flow `gh auth login -h garage.github.com`, the regular `github.com` token will be used for authenticating to garage environment.

PR #89, opened 2022-11-10
Overhaul README for stable release

mislav:


README.md
```diff
@@ -1,69 +1,61 @@
...
+# Go library for the GitHub CLI
 
-**This project is in beta!** Feedback is welcome.
+`go-gh` is a collection of Go modules to make authoring [GitHub CLI extensions][extensions] easier.
 
-A Go module for CLI Go applications and [gh extensions][extensions] that want a convenient way to interact with [gh][], and the GitHub API using [gh][] environment configuration.
+Modules from this library will obey GitHub CLI conventions by default:
 
-`go-gh` supports multiple ways of getting access to `gh` functionality:
+- [`CurrentRepository()`](https://pkg.go.dev/github.com/cli/go-gh#CurrentRepository) respects `GH_REPO` and reads from git remote configuration as fallback;
```
mntlty: nit: I don't often see semicolons at the end of a list item, it's usually either a `.` or empty

mntlty: Nit: 
```suggestion
- [`CurrentRepository()`](https://pkg.go.dev/github.com/cli/go-gh#CurrentRepository) respects the value of the `GH_REPO` environment variable if set, and reads from git remote configuration as fallback;
```

README.md
```diff
@@ -1,69 +1,61 @@
...
+- GitHub API requests will be authenticated using the same mechanism as gh, respecting `GH_TOKEN` and `GH_HOST` if present;
```
mntlty: Nit:
```suggestion
- GitHub API requests will be authenticated using the same mechanism as `gh`, respecting the value of the `GH_TOKEN` and `GH_HOST` environment variables if present;
```

README.md
```diff
@@ -1,69 +1,61 @@
...
+- Inspecting [`term`](https://pkg.go.dev/github.com/cli/go-gh/pkg/term) capabilities is done with respect to e.g. `NO_COLOR`;
```
mntlty: I didn't understand what this line meant 🤔

README.md
```diff
@@ -1,69 +1,61 @@
...
+- Generating [table-](https://pkg.go.dev/github.com/cli/go-gh/pkg/tableprinter) or [Go template](https://pkg.go.dev/github.com/cli/go-gh/pkg/template)-based output uses the same engine as gh;
```
mntlty: Nit: I think this is clearer without the dashes
```suggestion
- Generating [table](https://pkg.go.dev/github.com/cli/go-gh/pkg/tableprinter) or [Go template](https://pkg.go.dev/github.com/cli/go-gh/pkg/template) based output uses the same engine as gh;
```

README.md
```diff
@@ -1,69 +1,61 @@
...
+- The [`browser`](https://pkg.go.dev/github.com/cli/go-gh/pkg/browser) module activates the user's preferred web browser.
 
-* Helpers that automatically read a `gh` config to authenticate themselves
-* `gh.Exec` shells out to a `gh` install on your machine
-
-If you'd like to use `go-gh` on systems without `gh` installed and configured, you can provide custom authentication details to the `go-gh` API helpers.
+## Usage
 
+See the [go-gh full reference documentation](https://pkg.go.dev/github.com/cli/go-gh).
```
mntlty: Nit:
```suggestion
See the full `go-gh`  [reference documentation](https://pkg.go.dev/github.com/cli/go-gh) for more information
```
also suggest putting this with "See [examples](https://github.com/cli/go-gh/blob/8457ec84b80a60d0e06a01158258a53af04293e4/example_gh_test.go) for more demonstrations of usage." after the example code

README.md
```diff
@@ -1,69 +1,61 @@
-# go-gh beta
+# Go library for the GitHub CLI
```
mntlty, line 1: I think about this library as being generally useful, like a GitHub SDK in Go, and not only as a companion to the CLI for writing extensions. I suppose it comes down to how you want to market it!

samcoe (approved):
Thanks for taking the time to do this!

README.md
mislav: Good point, but our design for go-gh will always be influenced by GitHub CLI, and I wouldn't want to market go-gh as a general-purpose Go SDK for GitHub.

PR #92, opened 2022-12-13
Upgrade to Go 1.19

samcoe:
This PR upgrades Go to version 1.19.


mislav (approved):
Great, thank you. Ref. https://github.com/cli/cli/pull/6723

This is a backwards-incompatible change, but I don't think it's worth bumping the major version number. Do you think bumping just the minor version is fine?

samcoe:
@mislav I think a minor version bump is acceptable.

PR #95, opened 2023-01-13
V2

### LINKED ISSUE #97: GQLCLient should return GQLError for query and mutation methods
Right now the `query` and `mutations` methods on `GQLCLient` return an error type from [shurcooL-graphql](https://github.com/cli/shurcooL-graphql) which does not match the concrete error type from the `do` methods. This causes unnecessary complexity when trying to do type assertion on the errors returned from the `GQLCLient` methods. To fix this we can wrap the errors returned from `query` and `mutations` methods in a `GQLError`.

samcoe:
This PR introduces some API changes that are breaking and thus constitute bumping the major version to 2.0.0. This PR is best viewed by commits as each changes an individual package.

Changes by package:
* repository
    - Move `CurrentRepository` function from top level `gh` package to `repository` package. This felt like a more natural place for it.
    - Change name of `CurrentRepository` function to `Current` as it is now in the `repository` package.
    - Change `Repository` interface type to concrete type. The interface was not serving any purpose as there was only one implementation of it. Allowed for removing of `internal/repository/repository.go`.
* api
    - Change all references of `GQL` to `GraphQL`.
    - Move `RESTClient` function from top level `gh` package to `api` package. This made more sense in the `api` package than at the top level.
    - Move `GQLClient` function from top level `gh` package to `api` package. This made more sense in the `api` package than at the top level.
    - Move `HTTPClient` function from top level `gh` package to `api` package. This made more sense in the `api` package than at the top level.
    - Change `RESTClient` interface type to concrete type. The interface was not serving any purpose as there was only one implementation of it.
    - Change `GQLClient` interface type to concrete type. The interface was not serving any purpose as there was only one implementation of it.
    - Change `ClientOptions` to be used with value semantics instead of pointer semantics. This allows `ClientOptions` to be used in multiple clients without having to worry about them being modified.
    - Change methods on `HTTPError` and `GQLError` custom error types to take pointers as method receivers. This is a Go convention which should be followed unless good reason not to.
    - Added `DefaultRESTClient`, `DefaultHTTPClient`, `DefaultGQLClient` convenience functions. With the change to using value semantics for the options these felt nice to have.
    - These changes allowed for removing of `internal/api` folder.
    - Change `GQLClient` `Query` and `Mutate` methods to return `GQLError` instead of `shurcooL/graphql.Errors`.
* browser
    - Change `New` factory function to return `*Browser` instead of `Browser`. All methods on `Browser` use `*Browser` as the receiver so we should be returning a pointer.
* config
    - Change methods on `InvalidConfigFileError` and `KeyNotFoundError` custom error types to take pointers as method receivers. This is a Go convention which should be followed unless good reason not to.
* template
    - Change `New` factory function to return `*Template` instead of `Template`. All methods on `Template` use `*Template` as the receiver so we should be returning a pointer.

Fixes https://github.com/cli/go-gh/issues/97


pkg/repository/repository.go
```diff
@@ -3,88 +3,156 @@
...
+		return r, nil
 	default:
-		return nil, fmt.Errorf(`expected the "[HOST/]OWNER/REPO" format, got %q`, s)
+		return r, fmt.Errorf(`expected the "[HOST/]OWNER/REPO" format, got %q`, s)
+	}
+}
+
+// CurrentRepository uses git remotes to determine the GitHub repository
+// the current directory is tracking.
+func CurrentRepository() (Repository, error) {
```
samcoe: Perhaps this should be renamed to just `Current()` since it is now located in the `Repository` package?

vilmibm: I would vote for this as long as we are making a breaking change release.

vilmibm (approved):
thanks for all of this!

mislav (approved):
Thanks for doing the work! One tiny ask I would have is to use the full term `GraphQL` instead of `GQL`, since [GQL as a proposal](https://en.wikipedia.org/wiki/Graph_Query_Language) is a different query language than GraphQL.

PR #99, opened 2023-01-18
ExampleGQLClient_Mutate_simple

### LINKED ISSUE #98: Add a GraphQL Mutation with an `input` object to examples
I noticed that there are no examples of Mutations in https://github.com/cli/go-gh/blob/trunk/example_gh_test.go

I think this is important for a few reasons:
- there's no mention of [Mutate](https://github.com/cli/go-gh/blob/da64a13f361b2a1b9a1eb5358b7d7eee7506f6a2/pkg/api/client.go#L127) in the examples or in the readme. including it in the examples may make it more discoverable
- unlike queries, many of the GitHub GraphQL mutations take an input object, for example the [addStar mutation](https://docs.github.com/en/graphql/reference/mutations#addstar) and its [input](https://docs.github.com/en/graphql/reference/input-objects#addstarinput). because this is a nested object, it has a different format than queries such as https://github.com/cli/go-gh/blob/da64a13f361b2a1b9a1eb5358b7d7eee7506f6a2/example_gh_test.go#L125 Figuring out the right Go syntax took me some time even when I had a properly formed GraphQL mutation to compare it to

If you agree, I'd be happy to open a PR to update, and would like to know:
1. is `addStar` a good reference, or is there another that looks better?
2. for `variables`, would it be better to use https://github.com/shurcooL/githubv4 or to use only standard types?
3. should the Readme also mention mutations, or is it sufficient to have them in the examples?

mntlty:
Fixes https://github.com/cli/go-gh/issues/98

Adds a mutation for https://docs.github.com/en/graphql/reference/mutations#addstar with input and first params, and inline fragments on the response to demonstrate how these are used.

I tested this in a local `func main()` to confirm the format is correct and the request can succeed (output is `vilmibm`).

The node that global_id `R_kgDOF_MgQQ` references is left as an exercise for the user :smile:


example_gh_test.go
```diff
@@ -153,6 +154,43 @@ func ExampleGQLClient_advanced() {
...
+				Gist struct {
+					StargazerCount int
+					Url            string
+				} `graphql:"... on Gist"`
+			}
+		} `graphql:"addStar(input: $input)"`
+	}
+	variables := map[string]interface{}{
+		"input": githubv4.AddStarInput{
+			ClientMutationID: githubv4.NewString("addStar"),
```
mislav: Let's remove `ClientMutationID` since it's not necessary for the request

example_gh_test.go
```diff
@@ -153,6 +154,43 @@ func ExampleGQLClient_advanced() {
...
+	if err != nil {
+		log.Fatal(err)
+	}
+	var m struct {
+		AddStar struct {
+			Starrable struct {
+				Repository struct {
+					Stargazers struct {
+						Nodes []struct {
+							Login string
```
mislav: Why query the information about login names for the latest 10 stargazers? I don't feel that helps the example, and instead just adds noise. Maybe we could just fetch StargazerCount like for gists and display that as the result of the mutation

example_gh_test.go
```diff
@@ -153,6 +154,43 @@ func ExampleGQLClient_advanced() {
...
+						}
+					} `graphql:"stargazers(first: $first)"`
+				} `graphql:"... on Repository"`
+				Gist struct {
+					StargazerCount int
+					Url            string
```
mislav: The `Url` field should be called `URL` according to Go field naming conventions, but I'd say lets omit it unless we are going to consume the value

mislav (changes requested):
Looks great; thank you! Few nitpicks

example_gh_test.go
mntlty: my intent with ClientMutationID was to show how to convert inputs, but it looks like that works for StarrableID as well 😄

example_gh_test.go
mntlty: 👍

example_gh_test.go
mntlty: I have come across this somewhat regularly when using the apis, for example get the issues in a ProjectsV2 or the comments in a PullRequest, and wanted to showcase how that is used in a mutation. I do see that there are existing examples of it in `ExampleGQLClient_simple` and `ExampleGQLClient_advanced`, I'll remove it here

mntlty: oh - I also wanted to show the use of `first`, separate from `input`, which I don't see in other examples and is often required by the api. I removed it in a77f72b if you prefer without it, or we can revert it if you think it adds something

example_gh_test.go
```diff
@@ -164,31 +164,24 @@ func ExampleGQLClient_Mutate_simple() {
...
-					Url            string
 				} `graphql:"... on Gist"`
 			}
 		} `graphql:"addStar(input: $input)"`
 	}
 	variables := map[string]interface{}{
 		"input": githubv4.AddStarInput{
-			ClientMutationID: githubv4.NewString("addStar"),
-			StarrableID:      "R_kgDOF_MgQQ",
+			StarrableID: githubv4.NewString("R_kgDOF_MgQQ"),
```
mislav: Nit: for corectness, should this be `githubv4.NewID` instead of NewString? I don't think it makes a difference at runtime

mislav (approved):

example_gh_test.go
mntlty: @mislav agreed! changed in d7c93bdac5120e32f337ab747ccbc6334eee73ae

mislav:
Thank you!

Issue #23, opened 2022-01-28
Allow RESTClient and GQLClient to support automatic pagination


samcoe:
Pagination is currently not possible using `RESTClient` and takes lots of manual work using `GQLClient`. We should support this feature for both. Initial implementation idea is to add a `pagination` options to `ClientOptions` which would enable this for both clients and be a noop for the `HTTPClient`.

heaths:
Similar to cli/cli#1268. Perhaps there's a common solution for both. What @mislav [suggested](https://github.com/cli/cli/issues/1268#issuecomment-905383018) is a good idea. I would think with REST an array could simply be amended with subsequent elements. GraphQL might be more tricky.

PR #102, opened 2023-02-13
Add pagination examples for RESTClient and GQLClient

### LINKED ISSUE #100: Add pagination examples
We should add pagination examples for the REST and GQL APIs

samcoe:
Fixes https://github.com/cli/go-gh/issues/100

Additionally I removed our one reference to `github.com/shurcooL/githubv4`, it seemed like a large dependency to bring in just for an example.

### DIFF
--- a/example_gh_test.go
+++ b/example_gh_test.go
@@ -1,10 +1,13 @@
 package gh_test
 
 import (
+	"encoding/json"
 	"fmt"
 	"io"
 	"log"
+	"net/http"
 	"os"
+	"regexp"
 	"time"
 
 	gh "github.com/cli/go-gh"
@@ -12,7 +15,6 @@ import (
 	"github.com/cli/go-gh/pkg/tableprinter"
 	"github.com/cli/go-gh/pkg/term"
 	graphql "github.com/cli/shurcooL-graphql"
-	"github.com/shurcooL/githubv4"
 )
 
 // Execute 'gh issue list -R cli/cli', and print the output.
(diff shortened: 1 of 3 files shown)


example_gh_test.go
```diff
@@ -70,29 +72,66 @@ func ExampleRESTClient_request() {
...
+	}
+	requestPath := "repos/cli/cli/releases"
+	page := 1
+	hasNextPage := true
+	for hasNextPage {
+		response, err := client.Request(http.MethodGet, requestPath, nil)
+		if err != nil {
+			log.Fatal(err)
+		}
+		body, err := io.ReadAll(response.Body)
```
mislav: Nit: More recently I started to prefer `json.Decoder` to decode a Reader right directly into the destination struct without having to first save it to an intermediate buffer. Maybe you can try it here to simplify the example?

example_gh_test.go
```diff
@@ -70,29 +72,66 @@ func ExampleRESTClient_request() {
...
+		return "", false
+	}
+	client, err := gh.RESTClient(nil)
+	if err != nil {
+		log.Fatal(err)
+	}
+	requestPath := "repos/cli/cli/releases"
+	page := 1
+	hasNextPage := true
+	for hasNextPage {
```
mislav: Style nit: this could be an endless loop, with an explicit `break` happening if `findNextPage()` returned false. Logic would be lightly simpler and you wouldn't have to keep a `hasNextPage` variable outside of the loop

example_gh_test.go
```diff
@@ -70,29 +72,66 @@ func ExampleRESTClient_request() {
...
 	}
-
 	fmt.Printf("Asset downloaded to %s\n", f.Name())
 }
 
+// Get releases from cli/cli repository using REST API with paginated results.
+func ExampleRESTClient_pagination() {
+	var linkRE = regexp.MustCompile(`<([^>]+)>;\s*rel="([^"]+)"`)
+	findNextPage := func(response *http.Response) (string, bool) {
+		for _, m := range linkRE.FindAllStringSubmatch(response.Header.Get("Link"), -1) {
```
mislav, line 98: Thought for the future: having to do this as an extension author feels a bit low-level and requires understanding of certain details in the GitHub API. I wonder if we could provide facilities to encapsulate pagination from go-gh in the future?

example_gh_test.go
```diff
@@ -172,16 +211,58 @@ func ExampleGQLClient_Mutate_simple() {
 			}
 		} `graphql:"addStar(input: $input)"`
 	}
+	type AddStarInput struct {
+		StarrableID string `json:"starrableId"`
+	}
 	variables := map[string]interface{}{
-		"input": githubv4.AddStarInput{
```
mislav, line 176: I'm a bit torn: on one hand I do appreciate that the new example has fewer dependencies, but I also did find value in this example demonstrating the use of the shurcooL library that ensures typed inputs for each GraphQL mutation. I would actually recommend the use of that library to API clients such as extensions. @mntlty: thoughts? https://github.com/cli/go-gh/pull/99

mntlty: I think there's two use cases of what the examples here are for. 
- To show how `go-gh` can be used
- To show how `go-gh` should be used

If it's the first, having another package when a `struct` will do the job isn't necessary. If it's the second, having `githubv4` here introduces it to users who will likely benefit from the structure and safety it provides. They don't have to use it, but it's a big help to have the compile time checks.

Personally I learned about `githubv4` when contributing to the CLI, which not everyone is fortunate enough to be able to do ❤️

example_gh_test.go
samcoe: Agreed https://github.com/cli/go-gh/issues/23

example_gh_test.go
samcoe: My thoughts on this are that as `go-gh` is meant to be used in other projects and all of our dependencies come with it so we should try hard to minimize them. If `go` had a way to mark a dependency for test or examples only then this wouldn't matter. This was a case I felt that having the dependency was not worth it. What I think would be a good compromise would be to add a comment to this example where the struct is defined saying that the user could alternatively use the `githubv4` package where these are already defined. What do people think about that idea?

mislav: @samcoe I like that compromise!

mislav (approved):

mntlty (commented):
wish this existed a few weeks ago when I was figuring this out on my own, glad to see it's here now!

PR #103, opened 2023-02-13
Expose stdin and stdout in term package

### LINKED ISSUE #96: Support for stdin and stderr in term package
Right now the `term` package exposes `stdout` through the `Out` method, it would be nice if `stdin` and `stderr` were also exposed through some methods.

samcoe:
Fixes https://github.com/cli/go-gh/issues/96

### DIFF
--- a/pkg/term/env.go
+++ b/pkg/term/env.go
@@ -16,7 +16,7 @@ import (
 type Term struct {
 	in           *os.File
 	out          *os.File
-	err          *os.File
+	errOut       *os.File
 	isTTY        bool
 	colorEnabled bool
 	is256enabled bool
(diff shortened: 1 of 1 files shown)


pkg/term/env.go
```diff
@@ -71,11 +75,21 @@ func FromEnv() Term {
...
+	return t.in
+}
+
 // Out is the writer writing to standard output.
 func (t Term) Out() io.Writer {
 	return t.out
 }
 
+// Err is the writer writing to standard error.
+func (t Term) Err() io.Writer {
```
mislav: Nit: `Err` is sometimes in Go a method that returns an `error` object, for example `Context.Err()`

To be more clear about the purpose of this writer, could we maybe call it `ErrOut`?

mislav (approved):

PR #104, opened 2023-02-15
Replace shurcool Errors in GQLError

samcoe:
Fixes https://github.com/cli/go-gh/issues/97

Note this is based on the v2 branch as I felt this also could be considered a breaking change and thought it belonged with the other changes that will go into v2. If we felt like making this a non-breaking change we could also allow the `GQLError` to actually `Wrap`/`Unwrap` the shurcool error we received.


pkg/api/gql_client.go
```diff
@@ -109,7 +110,22 @@ func (c *GQLClient) Do(query string, variables map[string]interface{}, response
...
+		for i, e := range gqlErrs {
+			items[i] = GQLErrorItem{
+				Message:    e.Message,
+				Locations:  e.Locations,
+				Path:       e.Path,
+				Extensions: e.Extensions,
+				Type:       e.Type,
+			}
+		}
+		err = &GQLError{items}
```
mislav, line 126: This change could easily be backwards-compatible if GQLError implemented `Unwrap() error` that returned the original `graphql.Errors` instance

mislav (approved):

PR #105, opened 2023-02-16
Match hostnames case-insensitively

mislav:
TODO:
- [ ] check if the change to FindEntry will affect alias resolution?


mislav:
To avoid affecting case-sensitive alias resolution, I ideally wanted to perform case-insensitive lookups only for keys under `hosts`, but doing so required more changes to the implementation I was comfortable with, and the overall approach still feels brittle. Shelving this for now and making a simple workaround in the `cli/cli` repo.

PR #107, opened 2023-02-20
Read gh token from gh proper if not found in other locations

mislav:
GitHub CLI is soon moving away from storing the token in plain text configuration file. It will prefer to read the token from system's keyring storage.

As an unfortunate result, extensions will start failing now that token isn't in the YAML file anymore. But, we also don't want individual extensions to interface with the keyring. Instead, extensions will now shell out to `gh config get oauth_token --host HOSTNAME` to obtain the token.

This introduces a dependency on gh proper to be in PATH, but I consider that to be a reasonable request for extensions.

Delegating to the `gh` executable to obtain the token allows us to fine-tune how GitHub CLI interfaces with keyring storage in the future without affecting extensions again.

### DIFF
--- a/pkg/auth/auth.go
+++ b/pkg/auth/auth.go
@@ -29,7 +29,7 @@ const (
 
 // TokenForHost retrieves an authentication token and the source of that token for the specified
 // host. The source can be either an environment variable, configuration file, or the system
-// keyring. In the latter case, this shells out to "gh config get" to obtain the token.
+// keyring. In the latter case, this shells out to "gh auth token" to obtain the token.
 //
 // Returns "", "default" if no applicable token is found.
 func TokenForHost(host string) (string, string) {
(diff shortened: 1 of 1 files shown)


pkg/auth/auth.go
```diff
@@ -25,11 +27,30 @@ const (
...
+// keyring. In the latter case, this shells out to "gh config get" to obtain the token.
+//
 // Returns "", "default" if no applicable token is found.
 func TokenForHost(host string) (string, string) {
+	if token, source := TokenFromEnvOrConfig(host); token != "" {
+		return token, source
+	}
+
+	if ghExe, err := safeexec.LookPath("gh"); err == nil {
+		c := exec.Command(ghExe, "config", "get", "oauth_token", "--host", host)
```
samcoe: Is there a reason not to use `gh auth token --hostname`? Seems like this is the exact case we created that command.

samcoe (commented):
This looks good. Just left one question for you and would be nice to add a test for this new behavior.

PR #108, opened 2023-02-27
Only search secure storage for auth token

samcoe:
Follow up to https://github.com/cli/go-gh/pull/107. This is a small optimization to only check the secure storage if auth token is not found in other locations.

### DIFF
--- a/pkg/auth/auth.go
+++ b/pkg/auth/auth.go
@@ -87,7 +87,7 @@ func tokenForHost(cfg *config.Config, host string) (string, string) {
 }
 
 func tokenFromGh(path string, host string) (string, string) {
-	cmd := exec.Command(path, "auth", "token", "--hostname", host)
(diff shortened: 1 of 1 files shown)


pkg/auth/auth.go
```diff
@@ -87,7 +87,7 @@ func tokenForHost(cfg *config.Config, host string) (string, string) {
 }
 
 func tokenFromGh(path string, host string) (string, string) {
-	cmd := exec.Command(path, "auth", "token", "--hostname", host)
+	cmd := exec.Command(path, "auth", "token", "--secure-storage", "--hostname", host)
```
mislav, line 90: @samcoe Could there be downsides on a go-gh release relying on an yet-unreleased gh feature?

samcoe: I don't see any issue as the errors are being properly handled. I tested it on one of my extensions and it worked out fine.

Doni108108108:
g

PR #110, opened 2023-03-07
Skip JSON formatting when indent is ""

### LINKED ISSUE #109: jsonpretty should not add new lines if not indenting
https://github.com/cli/go-gh/blob/09213117cf60973e9d69fb5201b2643e5f1ec11b/pkg/jsonpretty/format.go#L55

Seems if `indent == ""` when passed to `jsonpretty.Format`, output should not be formatted at all i.e., delimiters - `[]{},` - should not have a newline appended.

heaths:
Fixes #109

### DIFF
--- a/pkg/jsonpretty/format.go
+++ b/pkg/jsonpretty/format.go
@@ -19,6 +19,7 @@ const (
 )
 
 // Format reads JSON from r and writes a prettified version of it to w.
+// Pass "" to indent to skip all formatting.
 func Format(w io.Writer, r io.Reader, indent string, colorize bool) error {
 	dec := json.NewDecoder(r)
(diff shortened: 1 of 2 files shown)


mislav (commented):
Hi, I didn't consider what the behavior for blank indent value would be, but in my view a JSON prettifier should still add newlines regardless of indent, since the two concepts are unrelated. And, if the caller wanted to pass empty indent to disable prettifying, couldn't they just skip calling `Format()` altogether?

heaths:
Fair. In this scenario, I was expecting a compact form but at long as the source was compact already - and both from the service and when piped to another program it would be - you're right: there's no reason to call this.

PR #113, opened 2023-03-19
Properly handle closing of files after writing

samcoe:
Addresses two CodeQL warnings about properly handling the return value when closing a file that has been written to. 

Fixes https://github.com/cli/go-gh/security/code-scanning/13
Fixes https://github.com/cli/go-gh/security/code-scanning/14

### DIFF
--- a/internal/api/cache.go
+++ b/internal/api/cache.go
@@ -170,33 +170,39 @@ func (fs *fileStorage) read(key string) (*http.Response, error) {
 	return res, err
 }
 
-func (fs *fileStorage) store(key string, res *http.Response) error {
+func (fs *fileStorage) store(key string, res *http.Response) (storeErr error) {
 	cacheFile := fs.filePath(key)
 
 	fs.mu.Lock()
 	defer fs.mu.Unlock()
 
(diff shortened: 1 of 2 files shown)


mislav (changes requested):
Ditto other PR: could we just ignore the errors? I don't think the end user benefits from `file.Close()` error aborting their gh command
```go
defer func() {
  _ = f.Close()
}()
```

samcoe:
@mislav I think the CodeQL warnings are not exactly around having an unhandled error and more that since we are writing to these files the data might be cached until the file handle is closed, so the errors that happen here are most likely to be errors writing that cached data to the file which feel more important than just a standard file closing error. We do normal expose write errors to the user as they indicate a potential lose of data.

Note there are other use cases of `defer f.Close()` that I didn't fix up because the file is only being read from and not written to.

mislav (approved):
> so the errors that happen here are most likely to be errors writing that cached data to the file which feel more important than just a standard file closing error.

Good point. In that case I agree about config file write errors being properly handled.

Note that `fileStorage.store()` error is typically discarded by the caller, so it's not really important to properly handle `Close()` there. You could still do it for completeness

Issue #114, opened 2023-03-20
Feature Request: Add support to call the `gh` command interactively


stemar94:
I would like to call `gh` interactively in order to use the already existing survey prompt functionality that comes with the standard features of `gh`.
With just `gh.Exec` I'm not able to access the interactive mode.

samcoe:
@stemar94 Thanks for writing in. This seems like a nice feature, I am in favor of adding it.

vilmibm:
this makes perfect sense, I'm +1

mislav:
Completed in https://github.com/cli/go-gh/commit/026e976b60f91b35bd97a2d82914c24515dda127, thank you!

PR #115, opened 2023-03-20
feat: add gh.ExecInteractive()

stemar94:
For #114

### DIFF
--- a/gh.go
+++ b/gh.go
@@ -54,14 +54,7 @@ func run(path string, env []string, args ...string) (stdOut, stdErr bytes.Buffer
 	cmd := exec.Command(path, args...)
 	cmd.Stdout = &stdOut
 	cmd.Stderr = &stdErr
-	if env != nil {
-		cmd.Env = env
-	}
-	err = cmd.Run()
-	if err != nil {
-		err = fmt.Errorf("failed to run gh: %s. error: %w", stdErr.String(), err)
-		return
-	}
+	err = runCmd(cmd, env)
 	return
 }
 
@@ -70,15 +63,19 @@ func runInteractive(path string, env []string, args ...string) (err error) {
 	cmd.Stdin = os.Stdin
 	cmd.Stdout = os.Stdout
 	cmd.Stderr = os.Stderr
+	return runCmd(cmd, env)
+}
+
+func runCmd(cmd *exec.Cmd, env []string) (err error) {
 	if env != nil {
 		cmd.Env = env
 	}
 	err = cmd.Run()
 	if err != nil {
-		err = fmt.Errorf("failed to run gh. error: %w", err)
+		err = fmt.Errorf("gh execution failed: %w", err)
 		return
 	}
-	return
+	return nil
 }
 
 // RESTClient builds a client to send requests to GitHub REST API endpoints.
--- a/gh.go
+++ b/gh.go
@@ -34,6 +34,16 @@ func Exec(args ...string) (stdOut, stdErr bytes.Buffer, err error) {
 	return run(path, nil, args...)
 }
 
+// Exec gh command interactively with provided arguments.
+func ExecInteractive(args ...string) (err error) {
+	path, err := path()
+	if err != nil {
+		err = fmt.Errorf("could not find gh executable in PATH. error: %w", err)
+		return
+	}
+	return runInteractive(path, nil, args...)
+}
+
 func path() (string, error) {
 	return safeexec.LookPath("gh")
 }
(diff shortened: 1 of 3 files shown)


stemar94:
Could also be nice to **_not wrap_** the `error`, if it's a proper `exec.ExitError`.

samcoe:
@stemar94 This is looking good so far. What would be the reason not to wrap an `exec.ExitError`? Go provides mechanisms to unwrap it if the user wants to.

gh.go
```diff
@@ -34,6 +34,16 @@ func Exec(args ...string) (stdOut, stdErr bytes.Buffer, err error) {
 	return run(path, nil, args...)
 }
 
+// Exec gh command interactively with provided arguments.
+func ExecInteractive(args ...string) (err error) {
+	path, err := path()
+	if err != nil {
+		err = fmt.Errorf("could not find gh executable in PATH. error: %w", err)
```
mislav: I understand that this basically copies the error message from `Exec()`, but we could maybe improve this error message in both functions.

The original error from `exec.LookPath` (which `safeexec` uses under the hood) will be something like:
```
exec: "gh": executable file not found in $PATH
```
After being wrapped, the full message will be:
```
could not find gh executable in PATH. error: exec: "gh": executable file not found in $PATH
```
I don't think the extra explanation added by wrapping adds any value here. I'd vote to avoid wrapping this error message altogether (and in the `Exec()` function too).

gh.go
```diff
@@ -53,6 +63,22 @@ func run(path string, env []string, args ...string) (stdOut, stdErr bytes.Buffer
...
+	cmd := exec.Command(path, args...)
+	cmd.Stdin = os.Stdin
+	cmd.Stdout = os.Stdout
+	cmd.Stderr = os.Stderr
+	if env != nil {
+		cmd.Env = env
+	}
+	err = cmd.Run()
+	if err != nil {
+		err = fmt.Errorf("failed to run gh. error: %w", err)
```
mislav: I don't mind that the original ExitError is wrapped here. The extra context makes it clear that it was the `gh` command that failed, and not code from the extension itself. Perhaps we can just make it a lighter wrapper:
```suggestion
		err = fmt.Errorf("gh: %w", err)
```

gh.go
```diff
@@ -34,6 +34,16 @@ func Exec(args ...string) (stdOut, stdErr bytes.Buffer, err error) {
 	return run(path, nil, args...)
 }
 
+// Exec gh command interactively with provided arguments.
+func ExecInteractive(args ...string) (err error) {
```
mislav: Since this is meant to spawn a potentially longer-running process, this function could accept a `context.Context` instance so that the caller has an option to kill the gh process when it needs to.

mislav (changes requested):
Thanks for this!

gh.go
stemar94: Done

gh.go
stemar94: I went for fmt.Errorf("gh execution failed: %w", err) for both Execs.

gh.go
stemar94: Done, also added a test.

stemar94:
> @stemar94 This is looking good so far. What would be the reason not to wrap an `exec.ExitError`? Go provides mechanisms to unwrap it if the user wants to.

I like for the ExecInteractive to happen transparently, where I like to hand over responsibility to the actual `gh`.
So it would be nice, if I could easily forward the exit code aswell.

Without unwrapping:
```
if execErr := gh.ExecInteractive(); execErr != nil {
  if exitErr, ok := execErr.(*exec.ExitError); ok {
    os.Exit(exitErr.ExitCode())
  }
}
```

With unwrapping:
```
if err := gh.ExecInteractive(); err != nil {
  if execErr := errors.Unwrap(err); execErr != nil {
    if exiterr, ok := execErr.(*exec.ExitError); ok {
      os.Exit(exiterr.ExitCode())
    }
  }
}
```

Also what does the one layer of wrapping give us in value?

mislav:
> With unwrapping:
> 
> ```
> if err := gh.ExecInteractive(); err != nil {
>   if execErr := errors.Unwrap(err); execErr != nil {
>     if exiterr, ok := execErr.(*exec.ExitError); ok {
>       os.Exit(exiterr.ExitCode())
>     }
>   }
> }
> ```

Go callers that handle specific error types should neither assume that the error value directly returned is a specific type nor manually call `Unwrap` to get to the desired type. Instead, they should use `errors.As()` to match a specific type:

```go
if err := gh.ExecInteractive(); err != nil {
  var exitError *exec.ExitError
  if errors.As(err, &exitError) {
    os.Exit(exitError.ExitCode())
  }
}
```

This way, it doesn't matter whether `err` is literally a `*exec.ExitError` or a wrapped `*exec.ExitError`. It will just work.

gh.go
```diff
@@ -28,12 +29,26 @@ import (
 func Exec(args ...string) (stdOut, stdErr bytes.Buffer, err error) {
 	path, err := path()
 	if err != nil {
-		err = fmt.Errorf("could not find gh executable in PATH. error: %w", err)
+		// No need to wrap error from path(), as it already has an appropriate
+		// form.
```
mislav: There is no need for these comments in the final PR changeset

mislav (commented):
This looks good! Feel free to mark your PR as "ready" when you get the test right so we can proceed to the final review. Thank you

mislav (approved):
Pushed some tweaks, including a new ExecContext function (to complement `Exec`) and added support for the GH_PATH environment variable that the parent `gh` process can set to inform extensions about the absolute location of `gh` on disk, in case `gh` isn't in PATH.

PR #116, opened 2023-03-27
jq: add option to pretty print JSON output

mjpieters:
Adds a new function `EvaluateFormatted` to maintain backward compatibility.
Indenting output by 2 spaces matches jq output.

Needed for cli/cli#3994

### DIFF
--- a/pkg/jq/jq.go
+++ b/pkg/jq/jq.go
@@ -2,6 +2,7 @@
 package jq
 
 import (
+	"bytes"
 	"encoding/json"
 	"fmt"
 	"io"
@@ -9,11 +10,24 @@ import (
 	"os"
 	"strconv"
 
+	"github.com/cli/go-gh/pkg/jsonpretty"
+
 	"github.com/itchyny/gojq"
 )
 
 // Evaluate a jq expression against an input and write it to an output.
+// Any top-level scalar values produced by the jq expression are written out
+// directly, as raw values and not as JSON scalars, similar to how jq --raw
+// works.
 func Evaluate(input io.Reader, output io.Writer, expr string) error {
+	return EvaluateFormatted(input, output, expr, "", false)
+}
+
+// Evaluate a jq expression against an input and write it to an output,
+// optionally with indentation and colorization.  Any top-level scalar values
+// produced by the jq expression are written out directly, as raw values and not
+// as JSON scalars, similar to how jq --raw works.
+func EvaluateFormatted(input io.Reader, output io.Writer, expr string, indent string, colorize bool) error {
 	query, err := gojq.Parse(expr)
 	if err != nil {
 		return err
@@ -39,7 +53,20 @@ func Evaluate(input io.Reader, output io.Writer, expr string) error {
 		return err
 	}
 
-	enc := json.NewEncoder(output)
+	var enc *json.Encoder
+	var buff bytes.Buffer
+	if !colorize {
+		// write straight to the output
+		// we can't use jsonpretty here because it handles indent = ""
+		// differently from json.Encoder (delimiters would be put on
+		// separate lines)
+		enc = json.NewEncoder(output)
+		enc.SetIndent("", indent)
+	} else {
+		// write to a buffer, and and then have jsonpretty format from this
+		buff = bytes.Buffer{}
+		enc = json.NewEncoder(&buff)
+	}
 
 	iter := code.Run(responseData)
 	for {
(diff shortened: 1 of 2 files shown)


pkg/jq/jq.go
```diff
@@ -14,6 +14,12 @@ import (
 
 // Evaluate a jq expression against an input and write it to an output.
 func Evaluate(input io.Reader, output io.Writer, expr string) error {
+	return EvaluateFormatted(input, output, expr, false)
+}
+
+// Evaluate a jq expression against an input and write it to an output, either without
+// whitespace or with 2 spaces of indentation per level (pretty printed).
+func EvaluateFormatted(input io.Reader, output io.Writer, expr string, pretty_print bool) error {
```
mislav: I would propose bringing these arguments more in line with the `pkg/jsonpretty` module from this repository: accepting `indent string` and `colorize bool` arguments and formatting the JSON accordingly:
- indent would be passed on to `enc`,
- if set, the colorize argument would cause everything to be passed through `jsonpretty`, which would write colorized, indented JSON to `output`.

BTW, Go naming conventions go against naming a variable in snake case like `pretty_print`. Instead, this would be typically named `prettyPrint`.

mislav (changes requested):
Thanks! This looks good, but I would propose one addition that would make this more flexible to use.

pkg/jq/jq.go
```diff
@@ -14,6 +14,12 @@ import (
 
 // Evaluate a jq expression against an input and write it to an output.
 func Evaluate(input io.Reader, output io.Writer, expr string) error {
+	return EvaluateFormatted(input, output, expr, false)
+}
+
+// Evaluate a jq expression against an input and write it to an output, either without
+// whitespace or with 2 spaces of indentation per level (pretty printed).
```
mislav: Forgot to add a request: the code docs of the original Evaluate function didn't specify in which cases this prints JSON payloads and in which raw values, but maybe your change is a good time to add this info, if you'd be willing?

Basically, this function takes a jq expression to extract and print raw (i.e. non-JSON) values to an output writer. However, if the selected node is not a scalar, but is instead a JSON object or an array, this function will print the resulting JSON. This is a hack meant not really to be consumed by scripts directly, but to offer feedback to the person crafting the jq expression, prompting them to potentially drill deeper to isolate all the scalar values. If the docs could make this clear, that would be great; thanks!

pkg/jq/jq.go
mjpieters: Right, I had made the assumption that you wouldn't want to duplicate the colorization functionality already present in the cli packages.

If this interface adds a `colorize` flag, then jq-processed data is colorized *here* (presumably) but non-jq-transformed data is colorized there still. My PR there routes the output from this package to be colorized by the same code path as non-jq JSON data.

Thanks for the naming issue pointer. It's a Python muscle-memory tripping hazard I run into more often, I'll correct it.

pkg/jq/jq.go
mjpieters: Sure, no problem. I had assumed that the code was designed to act as if `jq --raw` was being used (which does exactly this, output scalar values directly and not as JSON scalars).

BTW: I have used `gh ... --jq` plenty of times to output JSON as input to other tools. And I've yet to find a use-case where I'd want to output scalars and still have those JSON-formatted, so this has so far worked perfectly every single time. Except for the lack of pretty-printing of course, which makes debugging your jq expressions just that much harder. ;-)

pkg/jq/jq.go
mjpieters: I've now switched this to a `indent string, colorize bool` set of arguments, and delegate colorization to the jsonpretty package.

Note that this can _change behaviour_ when you set the indentation to the empty string but colorize to true. Where the stdlib json.Encoder outputs `[17,42]\n` in that case, jsonpretty outputs `[\n17,42\n]\n` (plus ANSI codes), putting a newline after every opening delimiter and before every closing delimiter. That might be considered a bug. At any rate, when colorize is false I don't use jsonpretty package to avoid precisely this issue.

mjpieters:
@mislav: thanks for the update there, it shows how green I am around my GoLang ears here.

mislav:
No worries and you're welcome. Your solution wasn't perfectly fine. My approach ins't necessarily less code but I think it might be easier to read and comprehend.

mislav (approved):
🚀

mjpieters:
> No worries and you're welcome. Your solution wasn't perfectly fine. My approach ins't necessarily less code but I think it might be easier to read and comprehend.

Well, I care a lot about readability, and the whole encoder setup was not giving me the 'this code just _fits_' feeling yet. I like the approach you took because that is a lot more readable. I've made note, so thanks again!

PR #119, opened 2023-06-01
jq: bump up gojq to fix printing empty arrays

itchyny:
## Summary
This PR fixes https://github.com/cli/cli/issues/7001. Not just `unique` filter, but also `flatten`, `group_by`, `unique_by`, `nth`, `indices`, `path`.

## History
There was an attempt to fix this issue at #101, but the behavior was rollbacked at https://github.com/cli/go-gh/pull/116/commits/c1803350d8c04d07bdd768aa471960d6f21e092b. This is essentially an unfriendly behavior of the gojq library, so I fixed gojq at https://github.com/itchyny/gojq/commit/96c4843a74198af45140b61528a895007e81862d with testing many array filters. I bumped up the version with a simple test updates to prevent future regression.


mislav (approved):
Thank you!

PR #120, opened 2023-06-02
Pin golangci-lint version

mislav:
Hopefully avoids flakey failure https://github.com/cli/go-gh/actions/runs/5150207133/jobs/9281631185?pr=119

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -21,6 +21,6 @@ jobs:
           git diff --exit-code go.mod
 
       - name: Lint
-        uses: golangci/golangci-lint-action@v3.1.0
+        uses: golangci/golangci-lint-action@v3.4.0
         with:
-          version: latest
+          version: v1.52


.github/workflows/lint.yml
```diff
@@ -21,6 +21,6 @@ jobs:
           git diff --exit-code go.mod
 
       - name: Lint
-        uses: golangci/golangci-lint-action@v3.1.0
+        uses: golangci/golangci-lint-action@v3.4.0
         with:
-          version: latest
+          version: v1.52
```
williammartin, line 26: Was there a reason not to pin this to the actual latest version?

williammartin: Is it cause they were released in the last 13 hours? 🤣

mislav: No reason. It's just that I have verified that this version works.

williammartin (approved):

PR #123, opened 2023-07-11
Export Path function

### LINKED ISSUE #121: Expose `ghLookupPath` or a better way to run `gh`
Been using `go-gh` for a few things and liking it so far!

I have a use case where I want to run `gh` as a long running command and I want to take over the writer for stdout and stderr. With the current APIs, I can't do this.

See example of how I am doing it right now by directly invoking gh: https://github.com/josebalius/gh-csfs/blob/main/internal/csfs/ssh.go#L37-L56

This solution works perfectly, but I am hardcoding `gh`, so it would be nice to either expose the lookup path function or a way to inject writers.

samcoe:
Closes https://github.com/cli/go-gh/issues/121

### DIFF
--- a/gh.go
+++ b/gh.go
@@ -18,7 +18,7 @@ import (
 
 // Exec invokes a gh command in a subprocess and captures the output and error streams.
 func Exec(args ...string) (stdout, stderr bytes.Buffer, err error) {
-	ghExe, err := ghLookPath()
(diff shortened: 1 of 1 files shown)


josebalius:
@vilmibm friendly ping on this as I'm waiting for it to land 😅 🙏

PR #126, opened 2023-08-04
add `CacheDir`

ffalor:
Adds `CacheDir` to be used in https://github.com/cli/cli/pull/7693


samcoe (commented):
This looks great so far, would you mind adding a simple test for `CacheDir()`?

samcoe (approved):

PR #128, opened 2023-08-11
fix(sanitizer): treat correct `U+FFFD` as normal character

### LINKED ISSUE #127: `asciisanitizer.Sanitizer` mishandled the `�` unicode character
We [have encountered an error](https://github.com/mdn/translated-content/pull/14909#issuecomment-1666099050) when using the GitHub `cli` to fetch commits in an MDN repository. And I found this error is coursed by the [sanitizer](https://github.com/cli/go-gh/blob/baae867137c67165ebd17e8ced4427fd9fcc44f5/pkg/asciisanitizer/sanitizer.go#L28-L97) which is used by GitHub `cli`.

So I created a [demo](https://github.com/cli/go-gh/files/12319924/src.zip) to reproduce the problem:

the plain text to transform:

```plain
�, plain text
```

When we read the plain text, and use `transform` with the the [sanitizer](https://github.com/cli/go-gh/blob/baae867137c67165ebd17e8ced4427fd9fcc44f5/pkg/asciisanitizer/sanitizer.go#L28-L97) , we would got an error:

image

But this should be the correct text. I found the error is returned [here](https://github.com/cli/go-gh/blob/baae867137c67165ebd17e8ced4427fd9fcc44f5/pkg/asciisanitizer/sanitizer.go#L50-L58).

So I read [the signature of `utf8.DecodeRune`](https://cs.opensource.google/go/go/+/refs/tags/go1.21.0:src/unicode/utf8/utf8.go;l=143-151). It may also return `utf8.RuneError` if the bytes are correctly decoded. And if there does be a decode error, it will return `(RuneError, 0)` or `(RuneError, 1)`.

So we can't judge whether there is a decoding error just based on the first value returned, like the text I used above, which uses this unicode character. The sanitizer mishandled it.

yin1999:
Fixes: #127.

As the length of `src` is larger than `1`, so the `utf8.DecodeRune()` will not return `(RuneError, 0)`, we can just check the following case: `(RuneError, 1)` to judge if there is a decode error.

### DIFF
--- a/pkg/asciisanitizer/sanitizer.go
+++ b/pkg/asciisanitizer/sanitizer.go
@@ -47,7 +47,7 @@ func (t *Sanitizer) Transform(dst, src []byte, atEOF bool) (nDst, nSrc int, err 
 			return
 		}
 		r, size := utf8.DecodeRune(src)
-		if r == utf8.RuneError {
+		if r == utf8.RuneError && size == 1 {
 			if !atEOF {
 				err = transform.ErrShortSrc
 				return
--- a/pkg/asciisanitizer/sanitizer.go
+++ b/pkg/asciisanitizer/sanitizer.go
@@ -47,7 +47,7 @@ func (t *Sanitizer) Transform(dst, src []byte, atEOF bool) (nDst, nSrc int, err 
 			return
 		}
 		r, size := utf8.DecodeRune(src)
-		if r == utf8.RuneError && size == 1 {
+		if r == utf8.RuneError && size < 2 {
 			if !atEOF {
 				err = transform.ErrShortSrc
 				return


pkg/asciisanitizer/sanitizer.go
```diff
@@ -47,7 +47,7 @@ func (t *Sanitizer) Transform(dst, src []byte, atEOF bool) (nDst, nSrc int, err
 			return
 		}
 		r, size := utf8.DecodeRune(src)
-		if r == utf8.RuneError {
+		if r == utf8.RuneError && size == 1 {
```
samcoe:
```suggestion
		if r == utf8.RuneError && size < 2 {
```

If the size is 0 or 1 there is an issue so lets check both cases.

samcoe (changes requested):
Thanks for getting started on this fix, I left one comment that I would like to see addressed.

yin1999:
> Thanks for getting started on this fix, I left one comment that I would like to see addressed.

Thanks for the suggestion. I've applied this :)

samcoe (approved):

PR #131, opened 2023-08-22
Implement prompter package

samcoe:
This PR implemented the prompter package which was mostly taken from `gh` with some small tweaks. Additionally, it includes a `PrompterMock` for use in tests.

cc https://github.com/cli/cli/issues/7898


pkg/prompter/prompter.go
```diff
@@ -0,0 +1,134 @@
...
+		return nil
+	}
+	return fmt.Errorf("could not prompt: %w", err)
+}
+
+// LatinMatchingFilter returns whether the value matches the input filter.
+// The strings are compared normalized in case.
+// The filter's diactritics are kept as-is, but the value's are normalized,
+// so that a missing diactritic in the filter still returns a result.
+func latinMatchingFilter(filter, value string, index int) bool {
```
andyfeller: (nit) Not sure if it matters but will match the function case just in case

```suggestion
// latinMatchingFilter returns whether the value matches the input filter.
// The strings are compared normalized in case.
// The filter's diactritics are kept as-is, but the value's are normalized,
// so that a missing diactritic in the filter still returns a result.
func latinMatchingFilter(filter, value string, index int) bool {
```

pkg/text/text.go
```diff
@@ -84,3 +88,19 @@ func RelativeTimeAgo(a, b time.Time) string {
 
 	return fmtDuration(int(ago.Hours()/24/365), "year")
 }
+
+// RemoveDiacritics returns the input value without "diacritics", or accent marks.
```
andyfeller, line 92: TIL

pkg/prompter/mock.go
```diff
@@ -0,0 +1,196 @@
...
+// use in tests. PrompterMock has a superset of the methods on Prompter
+// so they both can satisfy the same interface.
+type PrompterMock struct {
+	t                *testing.T
+	selectStubs      []selectStub
+	multiSelectStubs []multiSelectStub
+	inputStubs       []inputStub
+	passwordStubs    []passwordStub
+	confirmStubs     []confirmStub
+}
```
andyfeller: Assuming CLI extension integrators would use this in their own testing, how much documentation based on the [`cli/cli` usage](https://github.com/search?q=repo%3Acli/cli%20RegisterSelect&type=code) should be captured in documentation?

Do we expect people to depend upon https://pkg.go.dev/github.com/cli/go-gh or use markdown in the repo?

andyfeller (changes requested):
Comparing these changes versus https://github.com/cli/cli/tree/trunk/internal/prompter, this is a welcome addition!

Minor nit but also a larger question regarding documentation. 🙇

pkg/prompter/prompter.go
samcoe: Yeah this is something that I don't like about godoc comments... they encourage/enforce sentences but also the first letter can be uncapitalized, it bothers me.

pkg/prompter/mock.go
samcoe: Yeah I wanted to write up an example that compiles like the `ExamplePrompter()` but it started getting complicated due to the fact that I was trying to write a test about writing a test. Perhaps just a normal comment is enough. I will think on it a bit more.

samcoe: I went ahead and added a test example in the comments.

andyfeller (approved):

PR #141, opened 2023-10-18
Support migrations of the config

williammartin:
### Description

This PR relates to adding multi-account support to the CLI. I am opening it for discussion, not because I am confident is done and ready to merge, though I feel relatively good about the approach.

There are two parts to this PR one:
 * Adding [behaviour into the `config` package to support migrations
](https://github.com/cli/go-gh/compare/wm/multi-account-migration?expand=1#diff-a3d824da3c42420cd5cbb0a4a2c0e7b5bfddd819652788a0596d195dc6e31fa5R370)
 * Adding a migration that actually does the config migration

I feel like this is a reasonable split of responsibilities, rather than pushing it all into `go-gh/config` (which actually tries to be structure agnostic) or pulling it all up to `cli/cli` (which doesn't know anything about where it would store a backup for example).
 
I've left comments close to the places that I think are interesting, rather than trying to describe it all here.


go.mod
```diff
@@ -15,7 +15,7 @@ require (
 	github.com/mgutz/ansi v0.0.0-20200706080929-d51e80ef957d
 	github.com/muesli/reflow v0.3.0
 	github.com/muesli/termenv v0.13.0
-	github.com/stretchr/testify v1.7.0
+	github.com/stretchr/testify v1.8.4
```
williammartin, line 18: Required to use `require.EqualError`

pkg/config/config.go
```diff
@@ -139,29 +144,65 @@ func ReadFromString(str string) *Config {
...
+	return func(wo *WriteOpts) {
+		wo.IsBackup = true
+	}
+}
+
 // Write gh configuration files to the local file system.
 // It will only write gh configuration files that have been modified
 // since last being read.
-func Write(c *Config) error {
+func Write(c *Config, optFns ...WriteOpt) error {
```
williammartin: I'm not a fan of using this pattern here but it seemed like the quickest way to achieve the following:

```
When I want to write a backup of a config, I want it to be persisted regardless of whether anything was modified or not.
```

Unfortunately, it also means that we can't _only_ back up the bits that have changes (config.yml vs hosts.yml) easily. See more on a later comment about other approaches I explored.

pkg/config/config.go
```diff
@@ -320,3 +361,46 @@ http_unix_socket:
 # What web browser gh should use when opening URLs. If blank, will refer to environment.
 browser:
 `
+
+//go:generate moq -rm -out migration_mock.go . Migration
+type Migration interface {
+	Do(*Config) (bool, error)
+}
+
+func Migrate(c *Config, m Migration) error {
```
williammartin, line 347: So this is where `config` tries to actually support a migration. I split the migration support from the actual migration itself to try and tease apart the responsibilities. This function provides behaviour desired around a migration (avoiding broken config by taking a copy, not writing files when not required, and writing backup files) but leaves the actual migration to the injected `Migration`.

I made a couple of decisions here which I could go either way on, but felt reasonable:
 1. We take a copy of the entries to avoid destructive mutation on this `config`. In practice, any error bubbling out is likely to result in a complete CLI failure but I wanted to try and avoid any chance of us continuing with a broken config resulting in surprise behaviour (and definitely avoiding persisting it)
 2. I left it up to the migrations themselves to indicate whether a migration was required at all. An alternative would be to do a comparison of the entries, or to do an `IsModified` check but I kind of liked the explicitness.

williammartin, line 347: I played around with another approach that looked something like this:

```go
type Migration interface {
	Do(*Config) (bool, error)
}

type BackupConfigPersister interface {
    Persist(*Config) error
}

type ConfigPersister interface {
    Persist(*Config) error
}

type Migrator struct {
  m Migration
  bcp BackupConfigPersister
  cp ConfigPersister
}

func (m Migrator) Migrate(c *Config) error {
 ...
}
```

This would allow for custom behaviour between backup and config persisting and an obvious test seam. It's not bad but in practice it felt like it was a lot of abstraction for limited gain.

pkg/config/config.go
```diff
@@ -320,3 +361,46 @@ http_unix_socket:
...
+		return fmt.Errorf("failed to migrate config: %s", err)
+	}
+
+	// If there was no migration required, we're done
+	if !requiredMigration {
+		return nil
+	}
+
+	// Otherwise we'll write our current hosts config to a backup file
+	if err := writeAssumingLockHeld(c, asBackup()); err != nil {
```
williammartin: I made an attempt at writing some logic for "if only hosts.yml has changed then just back that up" but it seemed to add more complexity than it was worth to just avoid backing up the `config.yml` when it had no changes. 🤷

pkg/config/migration/multi_account.go
```diff
@@ -0,0 +1,124 @@
...
+//	 active_user: monalisa
+//     users:
+//	     monalisa:
+//		   git_protocol: https
+//
+// The reason for this is that we can then add new users under a host.
+
+type MultiAccount struct{}
+
+func (m MultiAccount) Do(c *config.Config) (bool, error) {
```
williammartin: Originally this was implemented in the `config` package itself, leaning on the private `yamlmap` entries. I decided to rewrite it to be agnostic of that, because it meant that either our migration was very explicitly private to `go-gh` or we would expose the inner yamlmap like `func Migrate(entries *yamlmap.Map)` which felt gross.

In practice, I actually think this version reads much nicer and simpler than the `yamlmap` version.

pkg/config/migration/multi_account.go
```diff
@@ -0,0 +1,124 @@
...
+		if err != nil {
+			return false, CowardlyRefusalError{fmt.Sprintf("couldn't get user name for %q", hostname)}
+		}
+
+		for _, configEntryKey := range configEntryKeys {
+			// We would expect that these keys map directly to values
+			// e.g. [williammartin, https, vim] but it's possible that a manually
+			// edited config file might nest further but we don't support that.
+			//
+			// We could consider throwing away the nested values, but I suppose
```
williammartin, line 107: Want to draw attention to this decision. We could also support it but it would certainly complicate the migration.

pkg/config/migration/multi_account.go
```diff
@@ -0,0 +1,124 @@
...
+//
+// github.localhost:
+//	 active_user: monalisa
+//     users:
+//	     monalisa:
+//		   git_protocol: https
+//
+// The reason for this is that we can then add new users under a host.
+
+type MultiAccount struct{}
```
williammartin, line 59: So, I put this down in the `config` package so it could live with this PR but I'm actually in two minds about where it belongs. The `config` exposed by `go-gh` is agnostic about its contents _except_ for the `hosts` key. In some ways it would be nice to abstract that knowledge and use some other generalised state to indicate when a file should be split but that's a topic for another time and too much work anyway.

Beyond that, the contents of the configuration are an _application_ level concern. That's when we provide meaning to the contents via the `AuthConfig`, or otherwise have expectations that we can read and write specific entries. The result of this is that perhaps this migration actually belongs in `cli/cli`. The seam is already there to support this.

pkg/config/migration/multi_account.go
```diff
@@ -0,0 +1,124 @@
...
+		return false, CowardlyRefusalError{"couldn't get hosts configuration"}
+	}
+
+	// If there are no hosts then it doesn't matter whether we migrate or not,
+	// so lets avoid any confusion and say there's no migration required.
+	if len(hostnames) == 0 {
+		return false, nil
+	}
+
+	// If there is an active user set for the first host then we must have already
```
williammartin: Want to draw attention to this decision.

pkg/config/migration/multi_account.go
```diff
@@ -0,0 +1,124 @@
...
+	if _, err := c.Get(append(hostsKey, hostnames[0], "active_user")); err == nil {
+		return false, nil
+	}
+
+	// Otherwise let's get to the business of migrating!
+	for _, hostname := range hostnames {
+		configEntryKeys, err := c.Keys(append(hostsKey, hostname))
+		// e.g. [user, git_protocol, editor]
+		if err != nil {
+			return false, CowardlyRefusalError{fmt.Sprintf("couldn't get host configuration despite %q existing", hostname)}
```
williammartin: None of these errors are tested. I also made the decision to return `false` even though a migration was required because it's idiomatic that under error, the other return values are ignored anyway. Didn't want to provide meaning to that value in the case of error. 🤷

pkg/config/config_test.go
```diff
@@ -632,6 +634,140 @@ func TestDefaultGeneralEntries(t *testing.T) {
...
+}
+
+func TestMigrationBackupWriteError(t *testing.T) {
+	// Create a location for our tests to write,
+	// Note that using env var here makes these tests unparallelisable.
+	tempDir := t.TempDir()
+	t.Setenv("GH_CONFIG_DIR", tempDir)
+
+	// Given we error when writing files (because we explicitly disallow it via this chmod)
+	require.NoError(t, os.Chmod(tempDir, 0000))
```
williammartin: This doesn't work on Windows apparently. Having an interface where we could inject faults on persistence operations would be handy but I think maybe it's just a lot of extra abstraction for limited gain right now.

pkg/config/config.go
samcoe: Feels like an simpler option would be to create a new function called `writeBackup` with almost duplicate logic as `Write` except we don't check if the config is modified and we write to `.bak` file.

pkg/config/config.go
samcoe: Both these decisions make sense to me.

pkg/config/config.go
```diff
@@ -320,3 +361,46 @@ http_unix_socket:
 # What web browser gh should use when opening URLs. If blank, will refer to environment.
 browser:
 `
+
+//go:generate moq -rm -out migration_mock.go . Migration
+type Migration interface {
+	Do(*Config) (bool, error)
```
samcoe: This function signature feels a bit odd to me, in that the `bool` return value is not indicating weather the migration was successful or not, it is meant to indicate if the migration was necessary at all. I wonder if splitting this into separate functions `IsNecessary` and `Do` would make this more clear. Obviously we can choose function names as we see fit.

pkg/config/config.go
samcoe: Agreed that this seems a bit overly complex.

pkg/config/config.go
samcoe: This feels perfectly reasonable 👍

pkg/config/config.go
```diff
@@ -320,3 +361,46 @@ http_unix_socket:
...
+	if err := writeAssumingLockHeld(copiedCfg); err != nil {
+		// Note that this error case is not tested as it was not easily possible to force an error
+		// here, and not on the backup write above, without introducing new abstractions which
+		// seemed to complicate the whole affair.
+		return fmt.Errorf("failed to write config after migration: %s", err)
+	}
+
+	// Finally, swap in the copied config entries last, since we know the file has been successfully
+	// persisted.
+	c.entries = copiedCfg.entries
```
samcoe, line 377: Do we actually want to do the swap or do we want to have this function return a pointer to the new migrated config? Not sure I have a strong opinion here, just a thought.

pkg/config/config_test.go
samcoe: We could do something like this to make it testable...

```go
func Migrate(c *Config, m Migration) error {
    return internalMigrate(c, m, writeFunc, writeBackupFunc)
}

// write tests with writing errors against internalMigrate instead of Migrate
func internalMigrate(c *Config, m Migration, w func(c *Config) error, wb func(c *Config) error) error {
    ... migration logic ...
}
```

Let me know if that is not clear. I don't have strong opinions on this either, just a lighter weight option than introducing an interface.

pkg/config/migration/multi_account.go
samcoe: I love the idea of putting this in `cli/cli` rather than in `cli/go-gh`.

pkg/config/migration/multi_account.go
samcoe: Wonderful!

pkg/config/migration/multi_account.go
samcoe: Makes sense to me. I am curious if we add in a version what places in `gh` would actually need to change? I can only think of maybe 1 or 2 off the top of my head.

pkg/config/migration/multi_account.go
samcoe: Perfectly reasonable to blow up here I think.

pkg/config/migration/multi_account.go
```diff
@@ -0,0 +1,124 @@
...
+//	 users:
+//	   williammartin:
+//	     git_protocol: https
+//	     editor: vim
+//
+// github.localhost:
+//	 active_user: monalisa
+//     users:
+//	     monalisa:
+//		   git_protocol: https
```
samcoe: Indentation here looks a bit off.

samcoe (commented):
Great start on this work! I left inline comments with my thoughts.

pkg/config/config.go
williammartin: I think that's fine. In various attempts I just duplicated the whole logic in line and left a comment saying "this was duplicated over there so if you change something here make sure to look over there too".

pkg/config/migration/multi_account.go
williammartin: Yeh so the question is "where does that version live?". Let's say it goes in `hosts.yml`, then it looks like this, which is fine:

```
version: v2
github.com:
 active_user: ...
ghe.io:
 active_user: ...
```

But the internal config structure is actually this:

```
...
hosts:
  version: 2
  github.com:
   active_user: ...
  ghe.io:
   active_user: ...
```

Now we have to special case anywhere that uses `c.Keys(..., "hosts")`.

On the other hand, if we put it in the config like:

```
...
hosts_version: 2
hosts:
  github.com:
   active_user: ...
  ghe.io:
   active_user: ...
```

Then we have a potential split brain between `config.yml` and `hosts.yml`.

In practice I think either work but they both leave me with a bit of 🤔 I'd really be in favour of having a version if we can. It makes future migrations a lot simpler.

pkg/config/config_test.go
williammartin: It is clear. I'm generally not a fan of testing the internals like this but since our tests already happen to live in the same package (rather than `_test`) maybe this is a reasonable approach. I'll noodle on it thanks.

pkg/config/migration/multi_account.go
williammartin: `gofmt` seems to be really unhappy with my attempts to put nested yaml in comments. I'll give it another go though.

pkg/config/config.go
williammartin: I think that's reasonable. It removes potential confusion about what the X means in `return X, err` as well.

pkg/config/migration/multi_account.go
samcoe: I would think putting it in the top level of `config.yml` is the right spot. Any change to the structure of `config.yml` or `hosts.yml` would trigger us bumping that version. We should treat them as a single config structure despite being in two separate files.

williammartin: Ooh, that's super obvious. I like it, and it makes me happy that I decided the `Migration` should work with the entire config rather than hardcoding it for just hosts.

pkg/config/migration/multi_account.go
samcoe: If `gofmt` is complaining I would say just leave it then... it is just a comment.

pkg/config/config_test.go
williammartin: Doing this approach didn't allow me control over the specific error paths (for hosts vs config) so I went with an integrated solution of making the files unwriteable.

pkg/config/config.go
williammartin: So the only downside of splitting this is the possibility for someone to call `Migrate` without checking first. Would adding godoc here be sufficient?

pkg/config/config.go
williammartin: I'm not sure if I understand you totally but I don't think we can return the migrated config because of the `sync.Once` package initialisation shenanigans. All places in the application point to the same config (unless someone has gone out of their way to avoid that).

pkg/config/config.go
samcoe: I think documentation here is sufficient 👍

pkg/config/config.go
samcoe: Hmm that is a good point around `sync.Once`, lets leave this as is then.

PR #142, opened 2023-10-24
Remove default configuration and add fallback configuration

samcoe:
This PR removes the knowledge of a default configuration from `go-gh` and replaces it with the ability to supply a fallback configuration to the `config.Read` function. This removes application level knowledge from `go-gh` and allows us to push it back into `gh`. One decision I made was that the fallback configuration would not be used unless both `config.yml` and `hosts.yml` were not found, previously the default configuration would be used whenever `config.yml` was not found. This change feels appropriate as we are now treating `config.yml` and `hosts.yml` as one single configuration and hides the details that they are actually two separate files from the caller.

⚠️ Note that this is a breaking change to the interface for `config.Read` which is exported, but in located in the `config` package which is only meant for use inside `gh` and has a warning about non-backwards compatible changes.

### DIFF
--- a/pkg/api/client_options.go
+++ b/pkg/api/client_options.go
@@ -89,7 +89,7 @@ func optionsNeedResolution(opts ClientOptions) bool {
 }
 
 func resolveOptions(opts ClientOptions) (ClientOptions, error) {
-	cfg, _ := config.Read()
+	cfg, _ := config.Read(nil)
 	if opts.Host == "" {
 		opts.Host, _ = auth.DefaultHost()
 	}
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -119,11 +119,15 @@ func (c *Config) Set(keys []string, value string) {
 	m.SetEntry(keys[len(keys)-1], yamlmap.StringValue(value))
 }
 
+func (c *Config) deepCopy() *Config {
+	return ReadFromString(c.entries.String())
+}
+
 // Read gh configuration files from the local file system and
-// return a Config. The fallback configuration will be returned
-// when there are no configuration files to load. If there are
-// no configuration files and no fallback configuration an empty
-// configuration will be returned.
+// returns a Config. A copy of the fallback configuration will
+// be returned when there are no configuration files to load.
+// If there are no configuration files and no fallback configuration
+// an empty configuration will be returned.
 var Read = func(fallback *Config) (*Config, error) {
 	once.Do(func() {
 		cfg, loadErr = load(generalConfigFile(), hostsConfigFile(), fallback)
@@ -185,7 +189,7 @@ func load(generalFilePath, hostsFilePath string, fallback *Config) (*Config, err
 		return nil, err
 	}
 
-	if generalMap == nil || generalMap.Empty() {
+	if generalMap == nil {
 		generalMap = yamlmap.MapValue()
 	}
 
@@ -203,8 +207,7 @@ func load(generalFilePath, hostsFilePath string, fallback *Config) (*Config, err
 	}
 
 	if generalMap.Empty() && fallback != nil {
-		//TODO: Do we need to copy fallback?
-		return fallback, nil
+		return fallback.deepCopy(), nil
 	}
 
 	return &Config{entries: generalMap}, nil
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -120,17 +120,18 @@ func (c *Config) Set(keys []string, value string) {
 }
 
 // Read gh configuration files from the local file system and
-// return a Config.
-var Read = func() (*Config, error) {
+// return a Config. The fallback configuration will be returned
+// when there are no configuration files to load. If there are
+// no configuration files and no fallback configuration an empty
+// configuration will be returned.
+var Read = func(fallback *Config) (*Config, error) {
 	once.Do(func() {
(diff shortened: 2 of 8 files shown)


pkg/config/config.go
```diff
@@ -201,6 +202,11 @@ func load(generalFilePath, hostsFilePath string) (*Config, error) {
 		generalMap.AddEntry("hosts", hostsMap)
 	}
 
+	if generalMap.Empty() && fallback != nil {
+		//TODO: Do we need to copy fallback?
```
samcoe: Open question here that could use some input.

pkg/config/config_test.go
```diff
@@ -274,36 +274,59 @@ func TestLoad(t *testing.T) {
...
 		t.Run(tt.name, func(t *testing.T) {
-			cfg, err := load(tt.globalConfigPath, tt.hostsConfigPath)
+			cfg, err := load(tt.globalConfigPath, tt.hostsConfigPath, tt.fallback)
 			if tt.wantErr {
 				assert.EqualError(t, err, tt.wantErrMsg)
 				return
 			}
 			assert.NoError(t, err)
+
+			fmt.Println(cfg)
```
williammartin: Looks like a leftover debug statement.

pkg/config/config_test.go
```diff
@@ -274,36 +274,59 @@ func TestLoad(t *testing.T) {
...
+
 			protocol, err := cfg.Get([]string{"git_protocol"})
-			assert.NoError(t, err)
-			assert.Equal(t, tt.wantGitProtocol, protocol)
+			if tt.wantGitProtocol == "" {
+				assert.EqualError(t, err, `could not find key "git_protocol"`)
```
williammartin: I kind of like:

```
var keyNotFoundError *KeyNotFoundError
assert.ErrorsAs(t, err, &keyNotFoundError)
```

String checks on errors are kinda last resort in my mind. That's why i introduced this helper:

```
func requireNoKey(t *testing.T, cfg *Config, keys []string) {
	t.Helper()

	_, err := cfg.Get(keys)
	var keyNotFoundError *KeyNotFoundError
	require.ErrorAs(t, err, &keyNotFoundError)
}
```

Also:

```
func requireKeyWithValue(t *testing.T, cfg *Config, keys []string, value string) {
	t.Helper()

	actual, err := cfg.Get(keys)
	require.NoError(t, err)

	require.Equal(t, value, actual)
}
```

Was handy in other cases.

pkg/config/config.go
```diff
@@ -185,7 +186,7 @@ func load(generalFilePath, hostsFilePath string) (*Config, error) {
 	}
 
 	if generalMap == nil || generalMap.Empty() {
```
williammartin: Is this `Empty()` check required anymore? Seems like it being empty would be equivalent to creating a new map?

pkg/config/config.go
williammartin: Yeh it's kinda funky because config isn't just a bag of data, it's actually a mutable structure. It's possible for the caller to end up modifying the fallback later without understanding the consequences. I had thought about providing a string before but that kind of leaked our internal config representation which would make it harder to change in the future.

I could be persuaded either way. Either we defensive copy and document to the caller that we are taking the values from the config, or we document that this **is** the Config struct that will be used as the fallback.

pkg/api/client_options.go
```diff
@@ -89,7 +89,7 @@ func optionsNeedResolution(opts ClientOptions) bool {
 }
 
 func resolveOptions(opts ClientOptions) (ClientOptions, error) {
-	cfg, _ := config.Read()
+	cfg, _ := config.Read(nil)
```
williammartin, line 92: It's kind of a shame with this pattern that we end up passing `nil` around everywhere. Oh well.

williammartin (changes requested):

pkg/api/client_options.go
samcoe: Agreed. Open to better suggestions if you have them.

williammartin: The only reasonable one I have is using variadic functional opts. Everything else is a much bigger change than I think we would want to tackle.

pkg/config/config.go
samcoe: I am going to change it to make a copy. That feels like the safer move.

pkg/config/config_test.go
```diff
@@ -274,36 +274,59 @@ func TestLoad(t *testing.T) {
...
+
+			if tt.wantToken == "" {
+				assertNoKey(t, cfg, []string{"hosts", "enterprise.com", "oauth_token"})
+			} else {
+				assertKeyWithValue(t, cfg, []string{"hosts", "enterprise.com", "oauth_token"}, tt.wantToken)
+			}
+
+			if tt.fallback != nil {
+				// Assert that load returns an equivalent copy of fallvback.
+				assert.Equal(t, tt.fallback.entries.String(), cfg.entries.String())
```
williammartin, line 327: Man, I totally get why the yaml Node has a slice for content now rather than a map, it's so it can always marshal in a stable order!

williammartin (approved):
I'm not sure there's ever a good reason to use `assert` instead of `require` but I don't think that should block this PR!

blg-666 (commented):
gh pr checkout 142

PR #143, opened 2023-10-25
Add support to config for setting and writing blank values

samcoe:
This PR updates the implementation of `config.Set` to set null yaml values instead of blank strings when passed a blank string as a value argument. We are not relying on the previous behavior anywhere and this allows us to create config keys even if they do not have any nested keys yet. We already work around this limitation in [auth login](https://github.com/cli/cli/blob/ce7d89de0cdc8f2e63021f2d2e252e3d7ff7f3b6/pkg/cmd/auth/login/login.go#L176) and we can finally remove that code. Additionally this is necessary for the work being done in this [branch](https://github.com/cli/cli/tree/wm/apply-multi-account-migration) which is not quiet ready for a PR yet.

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -102,7 +102,12 @@ func (c *Config) Remove(keys []string) error {
 // Set a string value in a Config.
 // The keys argument is a sequence of key values so that nested
 // entries can be set. If any of the keys do not exist they will
-// be created.
+// be created. If the string value to be set is empty it will be
+// represented as null not an empty string when written.
+//
+//	var c *Config
+//	c.Set([]string{"key"}, "")
+//	Write(c) // writes `key: ` not `key: ""`
 func (c *Config) Set(keys []string, value string) {
 	c.mu.Lock()
 	defer c.mu.Unlock()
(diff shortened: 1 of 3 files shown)


pkg/config/config.go
```diff
@@ -116,7 +121,11 @@ func (c *Config) Set(keys []string, value string) {
 		}
 		m = entry
 	}
-	m.SetEntry(keys[len(keys)-1], yamlmap.StringValue(value))
+	val := yamlmap.StringValue(value)
+	if value == "" {
+		val = yamlmap.NullValue()
+	}
+	m.SetEntry(keys[len(keys)-1], val)
```
andyfeller, line 128: Initially, I assumed we'd have the ability to set the config to null values, not that all empty strings will be null values.

I don't know a definitive way this will be a problem assuming this is YAML and should hopefully work 🤞

andyfeller (approved):

pkg/config/config.go
samcoe: It is YAML so they are more or less equivalent, which is why I went this direction instead of introducing a new function like `SetNull`. It felt mostly like implementation detail that the user shouldn't be concerned with. The user still receives an empty string using `Get` if they `Set` an empty string which felt like the only contract that we need to maintain here.

PR #145, opened 2023-11-20
Fix sanitization bug

samcoe:
Fixes https://github.com/cli/cli/issues/8336

There was a small bug in the logic of the sanitizer where multiple escaped ascii codes in a row would cause invalid JSON to be generated. This PR makes it so this does not happen.


andyfeller (approved):
LGTM!

PR #148, opened 2024-01-25
Support merging JSON arrays, objects

heaths:
Partly resolves cli/cli#1268 and replaces cli/cli#5652

### DIFF
--- a/pkg/jsonmerge/object.go
+++ b/pkg/jsonmerge/object.go
@@ -46,21 +46,14 @@ type objectMergerPage struct {
 	buffer bytes.Buffer
 }
 
+// Read caches the data in an internal buffer to be merged in Close.
+// No data is copied into p so it's not written to stdout.
 func (page *objectMergerPage) Read(p []byte) (int, error) {
-	// Read into a temporary buffer to be merged and written later.
-	p = make([]byte, len(p))
-	n, err := page.Reader.Read(p)
-	if err != nil {
-		return 0, err
-	}
-	if n == 0 {
-		return 0, io.EOF
-	}
-
-	_, err = page.buffer.Write(p[:n])
+	_, err := io.CopyN(&page.buffer, page.Reader, int64(len(p)))
 	return 0, err
 }
 
+// Close converts the internal buffer to a JSON object and merges it with the final JSON object.
 func (page *objectMergerPage) Close() error {
 	var src map[string]interface{}
 
--- a/pkg/jsonmerge/object.go
+++ b/pkg/jsonmerge/object.go
@@ -1,0 +1,73 @@
+package jsonmerge
+
+import (
+	"bytes"
+	"encoding/json"
+	"io"
+
+	"dario.cat/mergo"
+)
+
+type objectMerger struct {
+	io.Writer
+	dst map[string]interface{}
+}
+
+// NewObjectMerger creates a Merger for JSON objects.
+func NewObjectMerger(w io.Writer) Merger {
+	return &objectMerger{
+		Writer: w,
+		dst:    make(map[string]interface{}),
+	}
+}
+
+func (merger *objectMerger) NewPage(r io.Reader, isLastPage bool) io.ReadCloser {
+	return &objectMergerPage{
+		merger: merger,
+		Reader: r,
+	}
+}
+
+func (merger *objectMerger) Close() error {
+	// Marshal to JSON and write to output.
+	buf, err := json.Marshal(merger.dst)
+	if err != nil {
+		return err
+	}
+
+	_, err = merger.Writer.Write(buf)
+	return err
+}
+
+type objectMergerPage struct {
+	merger *objectMerger
+
+	io.Reader
+	buffer bytes.Buffer
+}
+
+func (page *objectMergerPage) Read(p []byte) (int, error) {
+	// Read into a temporary buffer to be merged and written later.
+	p = make([]byte, len(p))
+	n, err := page.Reader.Read(p)
+	if err != nil {
+		return 0, err
+	}
+	if n == 0 {
+		return 0, io.EOF
+	}
+
+	_, err = page.buffer.Write(p[:n])
+	return 0, err
+}
+
+func (page *objectMergerPage) Close() error {
+	var src map[string]interface{}
+
+	err := json.Unmarshal(page.buffer.Bytes(), &src)
+	if err != nil {
+		return err
+	}
+
+	return mergo.Merge(&page.merger.dst, src, mergo.WithAppendSlice, mergo.WithOverride)
+}
--- a/go.mod
+++ b/go.mod
@@ -3,6 +3,7 @@ module github.com/cli/go-gh/v2
 go 1.21
 
 require (
+	dario.cat/mergo v1.0.0
 	github.com/AlecAivazis/survey/v2 v2.3.7
 	github.com/MakeNowJust/heredoc v1.0.0
 	github.com/charmbracelet/glamour v0.6.0
--- a/go.sum
+++ b/go.sum
@@ -1,3 +1,5 @@
+dario.cat/mergo v1.0.0 h1:AGCNq9Evsj31mOgNPcLyXc+4PNABt905YmuqPYYpBWk=
+dario.cat/mergo v1.0.0/go.mod h1:uNxQE+84aUszobStD9th8a29P2fMDhsBdgRYvZOxGmk=
 github.com/AlecAivazis/survey/v2 v2.3.7 h1:6I/u8FvytdGsgonrYsVn2t8t4QiRnh6QSTqkkhIiSjQ=
 github.com/AlecAivazis/survey/v2 v2.3.7/go.mod h1:xUTIdE4KCOIjsBAE1JYsUPoCqYdZ1reCfTwbto0Fduo=
 github.com/MakeNowJust/heredoc v1.0.0 h1:cXCdzVdstXyiTqTvfqk9SDHpKNjxuom+DOlyEeQ4pzQ=
--- a/pkg/jsonmerge/array.go
+++ b/pkg/jsonmerge/array.go
@@ -1,0 +1,76 @@
+package jsonmerge
+
+import "io"
+
+type arrayMerger struct {
+	isFirstPage bool
+}
+
+// NewArrayMerger creates a Merger for JSON arrays.
+func NewArrayMerger() Merger {
+	return &arrayMerger{
+		isFirstPage: true,
+	}
+}
+
+func (merger *arrayMerger) NewPage(r io.Reader, isLastPage bool) io.ReadCloser {
+	return &arrayMergerPage{
+		merger:     merger,
+		Reader:     r,
+		isLastPage: isLastPage,
+	}
+}
+
+func (m *arrayMerger) Close() error {
+	// arrayMerger merges when reading, so any output was already written
+	// and there's nothing to do on Close.
+	return nil
+}
+
+type arrayMergerPage struct {
+	merger *arrayMerger
+
+	io.Reader
+	isLastPage bool
+
+	isSubsequentRead bool
+	cachedByte       byte
+}
+
+func (page *arrayMergerPage) Read(p []byte) (int, error) {
+	var n int
+	var err error
+
+	if page.cachedByte != 0 && len(p) > 0 {
+		p[0] = page.cachedByte
+		n, err = page.Reader.Read(p[1:])
+		n += 1
+		page.cachedByte = 0
+	} else {
+		n, err = page.Reader.Read(p)
+	}
+
+	if !page.isSubsequentRead && !page.merger.isFirstPage && n > 0 && p[0] == '[' {
+		if n > 1 && p[1] == ']' {
+			// Empty array case.
+			p[0] = ' '
+		} else {
+			// Avoid starting a new array and continue with a comma instead.
+			p[0] = ','
+		}
+	}
+
+	if !page.isLastPage && n > 0 && p[n-1] == ']' {
+		// Avoid closing off an array in case we determine we are at EOF.
+		page.cachedByte = p[n-1]
+		n -= 1
+	}
+
+	page.isSubsequentRead = true
+	return n, err
+}
+
+func (page *arrayMergerPage) Close() error {
+	page.merger.isFirstPage = false
+	return nil
+}
--- a/pkg/jsonmerge/array_test.go
+++ b/pkg/jsonmerge/array_test.go
@@ -1,0 +1,71 @@
+package jsonmerge
+
+import (
+	"bytes"
+	"io"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestArrayMerger_singleEmptyArray(t *testing.T) {
+	merger := NewArrayMerger()
+	w := &bytes.Buffer{}
+
+	r1 := bytes.NewBufferString(`[]`)
+	p1 := merger.NewPage(r1, true)
+	n, err := io.Copy(w, p1)
+	require.NoError(t, err)
+	assert.Equal(t, int64(2), n)
+	assert.NoError(t, p1.Close())
+	assert.Equal(t, `[]`, w.String())
+
+	require.NoError(t, merger.Close())
+}
+
+func TestArrayMerger_finalEmptyArray(t *testing.T) {
+	merger := NewArrayMerger()
+	w := &bytes.Buffer{}
+
+	r1 := bytes.NewBufferString(`["a","b"]`)
+	p1 := merger.NewPage(r1, false)
+	n, err := io.Copy(w, p1)
+	require.NoError(t, err)
+	assert.Equal(t, int64(8), n)
+	assert.NoError(t, p1.Close())
+	assert.Equal(t, `["a","b"`, w.String())
+
+	r2 := bytes.NewBufferString(`[]`)
+	p2 := merger.NewPage(r2, true)
+	n, err = io.Copy(w, p2)
+	require.NoError(t, err)
+	assert.Equal(t, int64(2), n)
+	assert.NoError(t, p2.Close())
+	assert.Equal(t, `["a","b" ]`, w.String())
+
+	require.NoError(t, merger.Close())
+}
+
+func TestArrayMerger_multiplePages(t *testing.T) {
+	merger := NewArrayMerger()
+	w := &bytes.Buffer{}
+
+	r1 := bytes.NewBufferString(`["a","b"]`)
+	p1 := merger.NewPage(r1, false)
+	n, err := io.Copy(w, p1)
+	require.NoError(t, err)
+	assert.Equal(t, int64(8), n)
+	assert.NoError(t, p1.Close())
+	assert.Equal(t, `["a","b"`, w.String())
+
+	r2 := bytes.NewBufferString(`["c","d"]`)
+	p2 := merger.NewPage(r2, true)
+	n, err = io.Copy(w, p2)
+	require.NoError(t, err)
+	assert.Equal(t, int64(9), n)
+	assert.NoError(t, p2.Close())
+	assert.Equal(t, `["a","b","c","d"]`, w.String())
+
+	require.NoError(t, merger.Close())
+}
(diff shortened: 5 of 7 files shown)


heaths:
@samcoe the idea here is to retain the basic - and intentionally rather minimal - design Nate put into place for JSON arrays and write them to stdout as we read them. This should have no practical impact to existing JSON array responses. Using the same interface - again, to retain the same basic design and not require near the level of refactoring cli/cli#5652 did - we can merge JSON objects using v1.0.0 of mergo, cache the merged object in memory, and write the merged object in the end. Given there can be multiple arrays and arrays will always be nested within at least the root `data` element, this seems the only way.

andyfeller:
@heaths : Thank you for this! ❤️  One thing I want to share is that @samcoe has left GitHub though hopefully will still be a contributor and advocate. 😞  Wanting to make sure we demonstrate our side of the commitment on this topic, I'm bringing this up with @williammartin so we can avoid the same stagnation from last time.

heaths:
Good luck to you, @samcoe, in your new venture! Thanks for all the support over the years!

andyfeller:
@heaths : having reviewed these changes a few times, I think I have a mental picture of how these changes fit into `gh api` but would like to confirm a few assumptions I'm making:

1. Moving the JSON merging logic here should primarily minimize the `gh api` changes to the [original `mergeJSON` logic](https://github.com/cli/cli/pull/5652/files#diff-86323aa9a70265152226ace11d3f3912e103103a5a82952bea0e6854de6fdab7R118-R148)

2. `mergeJSON` will still need to determine whether it is merging arrays or objects and use the appropriate `Merger`

Beyond that, is there anything else to consider regarding these changes and the original PR?

heaths:
No. In fact, I would pretty much forget the old PR exists. That was a huge refactor whereas you can see the new one on draft status is tiny. The logic of REST vs GraphQL was already determined, so I built on that. The goal was to keep the same basic flow that is currently in release: streaming for REST array responses (in case anyone took a dependency on that behavior), at least.

pkg/jsonmerge/object.go
```diff
@@ -0,0 +1,73 @@
...
+}
+
+func (page *objectMergerPage) Read(p []byte) (int, error) {
+	// Read into a temporary buffer to be merged and written later.
+	p = make([]byte, len(p))
+	n, err := page.Reader.Read(p)
+	if err != nil {
+		return 0, err
+	}
+	if n == 0 {
```
williammartin: I'm not sure about this conditional. I believe that a reader should only be returning (`0, nil`) when `len(p) == 0` as described in the `io.Reader` interface docs. However, with this conditional, in the case of `len(p) == 0` we're saying "there's no more to read" which isn't strictly correct.

Since we only seem to use this in `io.Copy` I'm not sure that we'll ever fall afoul of it but it also seems like something that might cause confusion?

pkg/jsonmerge/object.go
```diff
@@ -0,0 +1,73 @@
...
+type objectMergerPage struct {
+	merger *objectMerger
+
+	io.Reader
+	buffer bytes.Buffer
+}
+
+func (page *objectMergerPage) Read(p []byte) (int, error) {
+	// Read into a temporary buffer to be merged and written later.
+	p = make([]byte, len(p))
```
williammartin: Could this whole function be collapsed to:

```go
func (page *objectMergerPage) Read(p []byte) (int, error) {
	_, err := io.CopyN(&page.buffer, page.Reader, int64(len(p)))
	return 0, err
}
```

pkg/jsonmerge/object.go
```diff
@@ -0,0 +1,73 @@
...
+}
+
+type objectMergerPage struct {
+	merger *objectMerger
+
+	io.Reader
+	buffer bytes.Buffer
+}
+
+func (page *objectMergerPage) Read(p []byte) (int, error) {
```
williammartin, line 51: I think a doc comment here would be quite useful along the lines of:

"The `Read` function will never write into the provided buffer, it exists to copy from the page reader into the page buffer for later unmarshaling"

I went around a bit in circles wondering why we were always returning `0`.

pkg/jsonmerge/object.go
```diff
@@ -0,0 +1,73 @@
...
+
+// NewObjectMerger creates a Merger for JSON objects.
+func NewObjectMerger(w io.Writer) Merger {
+	return &objectMerger{
+		Writer: w,
+		dst:    make(map[string]interface{}),
+	}
+}
+
+func (merger *objectMerger) NewPage(r io.Reader, isLastPage bool) io.ReadCloser {
```
williammartin, line 24: It's a bit funky that the interface for `Merger` takes `isLastPage` when the object merger never uses it.

pkg/jsonmerge/object.go
```diff
@@ -0,0 +1,73 @@
...
+	_, err = page.buffer.Write(p[:n])
+	return 0, err
+}
+
+func (page *objectMergerPage) Close() error {
+	var src map[string]interface{}
+
+	err := json.Unmarshal(page.buffer.Bytes(), &src)
+	if err != nil {
+		return err
```
williammartin, line 62: Given that this can happen either because:
 * You didn't provide JSON
 * You didn't fully read the input buffer

Do you think it's worth first doing a check for whether there are still bytes to read in the input buffer and providing a useful error around that? Because we're using `io.Copy` in the api I'm not sure this can ever happen but maybe?

williammartin (commented):
Thanks for opening this @heaths, I have left a few comments and some thoughts inline here.

Firstly, I think you've done a really good job at minimising the changes required in the current `api` command code so kudos for that.

Secondly, though I don't think it should necessarily be done as part of this work, do you think it's worth exploring an earlier branching of `graphql` and `rest` in the API command? Things like `isFirstPage` and the streaming vs writing on `Close` behaviours are screaming that there's too much going on here. Similarly, when I look at the `api` command there's so many conditionals predicated on whether we're hitting graphql or not and it seems likely that would only increase. If we branched earlier, much of this PRs code would seem to become much simpler and idiomatic.

Finally, I see you put this in `go-gh`, is there a reason for that? Particularly since it seems tailored for our use case in the `api` command (e.g. `objectMergerPage.Read` not actually reading into the provided byte slice) I'm not sure that it makes sense for this to live as a consumable semvered package.

heaths:
> Secondly, though I don't think it should necessarily be done as part of this work, do you think it's worth exploring an earlier branching of graphql and rest in the API command? Things like isFirstPage and the streaming vs writing on Close behaviours are screaming that there's too much going on here. Similarly, when I look at the api command there's so many conditionals predicated on whether we're hitting graphql or not and it seems likely that would only increase. If we branched earlier, much of this PRs code would seem to become much simpler and idiomatic.

If you look at the original PR #5652 where I did refactor how a lot of this worked, it's huge. I'm not opposed to a major refactor similar to that, but I spent countless hours maintaining a clean merge since the PR was ignored for so long. I can't afford that again.

However, I don't think the idea here is wrong. Only when we need to paginate do we even need to consider merging, which is what these related changes do. Yes, the cli/cli does have a lot of branching on "/graphql" but the behavior for REST vs. GraphQL is very different, so that's not surprising. Paging is different, merging is different, output is different.

> Finally, I see you put this in go-gh, is there a reason for that? Particularly since it seems tailored for our use case in the api command (e.g. objectMergerPage.Read not actually reading into the provided byte slice) I'm not sure that it makes sense for this to live as a consumable semvered package.

The idea I discussed with @andyfeller and Sam was that because this was effectively a thin wrapper around `mergo` - while retaining Nate's "streaming" REST response merger - it could be useful to others. It's not unreasonable for extension authors to query for responses and may need to merge them as well. Having written several extensions myself - including a project V2 extension that called a lot of GraphQL queries and mutations - having utility functions like this in go-gh is very useful. That's why I've helped move a bunch of functionality or add some here. We could move it to cli/cli but I think it would be more valuable here.

I'll respond to individual comments.

pkg/jsonmerge/object.go
heaths: `io.Copy` never returns if we don't return `io.EOF`. It'll keep spinning. Tests were timing out, and jumping into the source of `io.Copy` showed why. But given this is effectively redirecting the response to a cache, telling any caller that nothing was read seems appropriate. The caller wouldn't have anything to read and the `io.Reader` truly would be at `io.EOF`.

pkg/jsonmerge/object.go
heaths: But the `arrayMerger` does, and it'll be different for each page. Interface implementations in any language won't always use every input. Any work to try-cast to different interfaces to make different calls doesn't seem worth the maintenance cost nor compute time given the limited scope. Feel free to reopen if you disagree.

pkg/jsonmerge/object.go
heaths: I will add a comment, but keep it mind it won't generate a doc comment. `objectMergerPage` is private. All you'd get for `Read` is from `io.ReadCloser.Read`.

pkg/jsonmerge/object.go
heaths: This wasn't intended to be go routine-safe. I'm not aware of much IO in the std library that is. By the time `Close` is called there should be no current or further calls to `Read` expected.

pkg/jsonmerge/object.go
williammartin: Gah, I'm sorry about this review comment. It was very lazily written. I meant to tie it back to the larger discussion of the consequences of the current API command layer implementation on this PR.

> Interface implementations in any language won't always use every input.

I wouldn't go so far as to say that this is never true but I do think that ignored arguments are generally trying to tell us something about the abstraction. In this case I think it's telling us that the calling code is forcing us to share an interface that would ideally be avoided. Yes, `arrayMerger` uses it because the REST pagination needs it, but it's only required on this interface because the calling code path is cramming graphql and REST together.

I don't think this interface is the problem, I think it's a symptom of a problem elsewhere.

I'm content to leave this resolved though.

pkg/jsonmerge/object.go
williammartin: I'm not talking about goroutines here. Here's a test that demonstrates what I mean:

```go
func TestObjectMerger_notFullyRead(t *testing.T) {
	w := &bytes.Buffer{}
	merger := NewObjectMerger(w)

	r1 := bytes.NewBufferString(`{}`)
	p1 := merger.NewPage(r1, true)
	n, err := p1.Read(make([]byte, 1)) // only reading 1 byte

	require.NoError(t, err)
	assert.Equal(t, 0, n)

	require.NoError(t, p1.Close())
}
```

```
--- FAIL: TestObjectMerger_notFullyRead (0.00s)
    /Users/williammartin/workspace/go-gh/pkg/jsonmerge/object_test.go:87: 
        	Error Trace:	object_test.go:87
        	Error:      	Received unexpected error:
        	            	unexpected end of JSON input
        	Test:       	TestObjectMerger_notFullyRead
```

This fails because we only `Read` 1 byte of the page and couldn't unmarshal a valid JSON document.

This leans into my concern about shipping this as its own package as-is because it's tightly coupled to the expectations of the calling code (i.e. that you consume every byte which we _know_ happens because we `io.Copy` but that's not required by someone calling `Read`) and that isn't documented anywhere on the package interface.

I'm suggesting in my review comment that if there are still bytes in the `Reader` when we call `Close`, we could help users of this package by pointing them to this fact, rather than having them scratching their head about the structure of the string they passed in. It seems like a nice quality of life improvement.

Yes, it's obvious that you need to ensure that the entire document is consumed to have a valid JSON structure but especially if you happen to be some degrees of nested different `Reader`s away it's not always obvious what's happening. And I know you're a Rust developer and value the experience of useful errors 😬 

---

Also, though I care less about this one, it's possible that an _invalid_ JSON string can be consumed and merged unexpectedly by not emptying the `Reader`:

```go
func TestObjectMerger_notFullyReadInvalidObject(t *testing.T) {
	w := &bytes.Buffer{}
	merger := NewObjectMerger(w)

	r1 := bytes.NewBufferString(`{}}`)
	p1 := merger.NewPage(r1, true)
	n, err := p1.Read(make([]byte, 2))

	require.NoError(t, err)
	assert.Equal(t, 0, n)

	require.Error(t, p1.Close())
}
```

Though I haven't looked as closely, I would assume the same is true for the `arrayMerger`.

pkg/jsonmerge/object.go
williammartin: I can't make any test in the cli (pointing at this PR branch) or `go-gh` fail without this line and I don't really see why it would. There's definitely no test  in `go-gh` which hits this code path. Can you point to the bit that fails without this?

Here's a test that demonstrates what I mean:

```go
func TestObjectMerger_ReadZeroThenSome(t *testing.T) {
	w := &bytes.Buffer{}
	merger := NewObjectMerger(w)

	r1 := bytes.NewBufferString(`{"a":1,"b":2}`)
	p1 := merger.NewPage(r1, true)

	var buf []byte
	for i := 0; i < 10; i++ {
		buf = make([]byte, i)
		_, err := p1.Read(buf)
		if err == io.EOF {
			break
		}
	}

	require.NoError(t, merger.Close())
	assert.JSONEq(t, `{"a":1,"b":2}`, w.String())
}
```

Functionally this is a `Copy` but starting from a byte slice size of `0`. Since `len(p) == 0`, the `BufferString.Read` function [dutifully returns `0`](https://cs.opensource.google/go/go/+/refs/tags/go1.21.6:src/bytes/buffer.go;l=323-325) as described in the `io.Reader` interface doc comment but we then treat that as `io.EOF` even though there is in fact the entire JSON string left to read.

In practice I don't know when this is an issue but given that it's explicitly called out in the `io.Reader` doc comment it makes me nervous doing this, especially since I don't understand what it's actually solving.

williammartin:
>  I'm not opposed to a major refactor similar to that, but I spent countless hours maintaining a clean merge since the PR was ignored for so long. I can't afford that again.

Just to be super clear, I have no expectations for this to happen:
 * As part of this work
 * On your back

I'm asking for your opinion as someone who has more familiarity with the code than Andy or me.

> However, I don't think the idea here is wrong

I also would not use the word wrong here. As in my comment above, you've done a remarkable job at providing something that explicitly doesn't require wholesale changes in the command layer. It's crafty!

> Yes, the cli/cli does have a lot of branching on "/graphql" but the behavior for REST vs. GraphQL is very different, so that's not surprising. Paging is different, merging is different, output is different.

The point I'm failing to make is that the API command branches 7 times (at a cursory count) based on whether this is a graphql request or not. That makes for a lot of state that needs to be carried around mentally when reading the code. Additionally, it forces other abstractions to contort into supporting the shared code path (see `isLastPage`, and `objectMergerPage` which says it satisfies the `io.Reader` interface but isn't really a valid implementation as per the interface doc comments because it never reads bytes into `p`).

If instead, we branched once, early, I'm curious how that tradeoff would play out. Yes, there would be some duplication but I wonder if we're in a case of ["duplication is far cheaper than the wrong abstraction"](https://sandimetz.com/blog/2016/1/20/the-wrong-abstraction). The more code that gets written to support this style, the more we will feel the pressure to continue to invest in it over time.

Again, I'm not asking for you to do any work to implement this, or even to validate it, I'm just interested in whether it resonates with you given your experience poking around in here. If you were to say "I don't know and I don't have the time to investigate it" that's totally understandable.

> Having written several extensions myself - including a project V2 extension that called a lot of GraphQL queries and mutations - having utility functions like this in go-gh is very useful.

Thanks for the insight. I'm not opposed to keeping it here though I have some reservations about the existing abstraction. Having it here doesn't require that we maintain it beyond keeping it exposed and not breaking it, and if it's useful in its current state so be it.

I've responded to individual comments and unresolved those that I would like a response to.

williammartin:
Also, just to set expectations, I don't think there is anything about this PR that is really blocking. I just want to make sure we all get on the same page about its usage and limitations. Right now I fully expect that the PR will not change significantly, and will get merged in shortly. I certainly don't intend to drag this out especially given the experience we talked about before on #5652.

pkg/jsonmerge/object.go
heaths: I see what you mean. But how would we ensure there's more to read? Just try reading from the underlying `io.Reader` again and make sure we get 0 bytes and/or `io.EOF`?

That said, I appreciate the concern about putting this in cli/go-gh and making it more "general purpose". I'm fine with porting it to cli/cli and keeping it internal. If ever there's a need - like many of the things including the `tableprinter` I helped improve and migrate out - to move it out, we can deal with generalizing it then.

Sound good? I could still make some improvements to better guard against possible issues, but making this internal and controlling the use patterns would mitigate several of these concerns.

heaths:
> If instead, we branched once, early, I'm curious how that tradeoff would play out. Yes, there would be some duplication but I wonder if we're in a case of ["duplication is far cheaper than the wrong abstraction"](https://sandimetz.com/blog/2016/1/20/the-wrong-abstraction). The more code that gets written to support this style, the more we will feel the pressure to continue to invest in it over time.
> 
> Again, I'm not asking for you to do any work to implement this, or even to validate it, I'm just interested in whether it resonates with you given your experience poking around in here. If you were to say "I don't know and I don't have the time to investigate it" that's totally understandable.

I see. Thanks for clarifying. Yes, bifurcating early might actually be better. That's why keeping my original PR clean against your upstream `trunk` was so expensive. As other PRs came in - often just changing a line or two - the branching was so complicated it would throw everything off. It was also hard to understand initially, which made my PR refactoring much of the code even harder to understand - even for me when cleaning up merge conflicts.

I'm open to doing this work. What is your timeline for the next release, though? I have a couple thoughts depending on that:

1. Close this PR and move it to the `internal` package in cli/cli. Make some of the changes we've discussed above (I haven't tried the `CopyN` suggestion yet but will once we settle on some other feedback).
2. Merge that for an initial implementation.
3. As a separate PR, refactor.

If we split the work, an initial release to merge JSON arrays and objects could ship. Refactoring later should have no user impact. That said, I'd also have less incentive to finish 3, but still interested. That said, if one of your team wants to work on that, I wouldn't be offended. :smile:

pkg/jsonmerge/object.go
williammartin: If we moved this to `cli/internal` then I would be comfortable not including this change since we know we're in control of the usage.

> Just try reading from the underlying io.Reader again and make sure we get 0 bytes and/or io.EOF?

That's what I would have tried to do, perhaps naively.

williammartin:
> I see. Thanks for clarifying. Yes, bifurcating early might actually be better. That's why keeping my original PR clean against your upstream trunk was so expensive. As other PRs came in - often just changing a line or two - the branching was so complicated it would throw everything off. 

Yeh understood. Total nightmare for everyone once that starts happening.

> What is your timeline for the next release, though? 

We release every 2 weeks or so, sometimes more frequently. I would anticipate we ship by next Tuesday at the latest. I don't think any large scale refactor would need to make it into the next release though.

I think we should proceed with getting this work into `cli/cli` pronto. This is a useful feature and we absolutely don't want to block it on some nebulous refactor that provides uncertain value and has an unknown time-bound. It may end up being more toil and risk than we are comfortable with.

> That said, I'd also have less incentive to finish 3, but still interested. That said, if one of your team wants to work on that, I wouldn't be offended. 😄

So just for context right now I'm the only full time maintainer on the core CLI. @andyfeller is a machine and jumps back and forth between this and copilot CLI work. We are looking to add more engineers but it's a process.

That said, I do think that I would find it valuable to get my hands grubby on this one since the refactor isn't time sensitive. There's a lot I still need to learn about the nuances of this command. I don't think I'd looked at the implementation of `gh api` before this PR and trying to offer constructive review without experience in the code is challenging.

pkg/jsonmerge/object.go
heaths: As discussed below in general comments, moving this to cli/cli as an internal package will suffice.

pkg/jsonmerge/object.go
heaths: Resolved with change to `io.CopyN`.

pkg/jsonmerge/object.go
heaths: This works, but I got to thinking: my original code is wrong, and it's not fixed by using `io.CopyN`. We may actually read less than `len(p)`.

That said, I supposed it doesn't matter given the docs:

> CopyN copies n bytes (or until an error) from src to dst. It returns the number of bytes copied and the earliest error encountered while copying. On return, written == n if and only if err == nil.

We'd get an error if `written != n`.

heaths:
Closing per discussion above and in cli/cli#8620.

PR #151, opened 2024-02-26
Treat tenancy as non enterprise

williammartin:
## Description

I think there's some things I'd like to move from https://github.com/cli/cli/blob/24481c5c2e49d3529a6621953b8878fe0514ff22/internal/ghinstance/host.go into `go-gh` but I think having this behaviour as an obvious standalone change is valuable without muddying the waters.

### DIFF
--- a/pkg/auth/auth.go
+++ b/pkg/auth/auth.go
@@ -3,6 +3,7 @@
 package auth
 
 import (
+	"fmt"
 	"os"
 	"os/exec"
 	"strconv"
@@ -148,8 +149,15 @@ func defaultHost(cfg *config.Config) (string, string) {
 	return github, defaultSource
 }
 
+// TenancyHost is the domain name of a tenancy GitHub instance.
+const tenancyHost = "ghe.com"
+
 func isEnterprise(host string) bool {
-	return host != github && host != localhost
+	return host != github && host != localhost && !isTenancy(host)
+}
+
+func isTenancy(host string) bool {
+	return strings.HasSuffix(host, "."+tenancyHost)
 }
 
 func normalizeHostname(host string) string {
@@ -160,5 +168,21 @@ func normalizeHostname(host string) string {
 	if strings.HasSuffix(hostname, "."+localhost) {
 		return localhost
 	}
+	// This has been copied over from the cli/cli NormalizeHostname function
+	// to ensure compatible behaviour but we don't fully understand when or
+	// why it would be useful here. We can't see what harm will come of
+	// duplicating the logic.
+	if before, found := cutSuffix(hostname, "."+tenancyHost); found {
+		idx := strings.LastIndex(before, ".")
+		return fmt.Sprintf("%s.%s", before[idx+1:], tenancyHost)
+	}
 	return hostname
+}
+
+// Backport strings.CutSuffix from Go 1.20.
+func cutSuffix(s, suffix string) (string, bool) {
+	if !strings.HasSuffix(s, suffix) {
+		return s, false
+	}
+	return s[:len(s)-len(suffix)], true
 }
--- a/pkg/auth/auth_test.go
+++ b/pkg/auth/auth_test.go
@@ -288,6 +288,16 @@ func TestNormalizeHostname(t *testing.T) {
 			host:     "mygithub.com",
 			wantHost: "mygithub.com",
 		},
+		{
+			name:     "bare tenant",
+			host:     "tenant.ghe.com",
+			wantHost: "tenant.ghe.com",
+		},
+		{
+			name:     "subdomained tenant",
+			host:     "api.tenant.ghe.com",
+			wantHost: "tenant.ghe.com",
+		},
 	}
 
 	for _, tt := range tests {
--- a/pkg/auth/auth_test.go
+++ b/pkg/auth/auth_test.go
@@ -247,6 +247,11 @@ func TestIsEnterprise(t *testing.T) {
 			host:    "mygithub.com",
 			wantOut: true,
 		},
+		{
+			name:    "tenant",
+			host:    "tenant.ghe.com",
+			wantOut: false,
+		},
 	}
 
 	for _, tt := range tests {
--- a/pkg/auth/auth_test.go
+++ b/pkg/auth/auth_test.go
@@ -84,6 +84,30 @@ func TestTokenForHost(t *testing.T) {
 			wantToken:  "yyyyyyyyyyyyyyyyyyyy",
 			wantSource: "oauth_token",
 		},
+		{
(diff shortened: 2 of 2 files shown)


pkg/auth/auth.go
```diff
@@ -148,8 +148,15 @@ func defaultHost(cfg *config.Config) (string, string) {
...
+const tenancyHost = "ghe.com"
+
 func isEnterprise(host string) bool {
-	return host != github && host != localhost
+	return host != github && host != localhost && !isTenancy(host)
+}
+
+func isTenancy(host string) bool {
+	return strings.HasSuffix(host, "."+tenancyHost)
 }
```
andyfeller, line 161: Any concerns about `cli/go-gh` not 1) normalizing the hostname as part of the boolean checks or 2) extracting the tenant name from a host name like [`cli/cli`](https://github.com/cli/cli/blob/24481c5c2e49d3529a6621953b8878fe0514ff22/internal/ghinstance/host.go#L23-L60)?

```go
// IsEnterprise reports whether a non-normalized host name looks like a GHE instance.
func IsEnterprise(h string) bool {
	normalizedHostName := NormalizeHostname(h)
	return normalizedHostName != defaultHostname && normalizedHostName != localhost
}

// IsTenancy reports whether a non-normalized host name looks like a tenancy instance.
func IsTenancy(h string) bool {
	normalizedHostName := NormalizeHostname(h)
	return strings.HasSuffix(normalizedHostName, "."+tenancyHost)
}

// TenantName extracts the tenant name from tenancy host name and
// reports whether it found the tenant name.
func TenantName(h string) (string, bool) {
	normalizedHostName := NormalizeHostname(h)
	return cutSuffix(normalizedHostName, "."+tenancyHost)
}

func isGarage(h string) bool {
	return strings.EqualFold(h, "garage.github.com")
}

// NormalizeHostname returns the canonical host name of a GitHub instance.
func NormalizeHostname(h string) string {
	hostname := strings.ToLower(h)
	if strings.HasSuffix(hostname, "."+defaultHostname) {
		return defaultHostname
	}
	if strings.HasSuffix(hostname, "."+localhost) {
		return localhost
	}
	if before, found := cutSuffix(hostname, "."+tenancyHost); found {
		idx := strings.LastIndex(before, ".")
		return fmt.Sprintf("%s.%s", before[idx+1:], tenancyHost)
	}
	return hostname
}
```

andyfeller (commented):
Would we want to bring over the [`isTenancy()` tests](https://github.com/cli/cli/blob/24481c5c2e49d3529a6621953b8878fe0514ff22/internal/ghinstance/host_test.go#L52-L89) over too?

williammartin:
> Would we want to bring over the [isTenancy() tests](https://github.com/cli/cli/blob/24481c5c2e49d3529a6621953b8878fe0514ff22/internal/ghinstance/host_test.go#L52-L89) over too?

Not specifically but this reminded me that I definitely want to add tests to https://github.com/cli/go-gh/blob/d88d88f917cd4e0bcaa582bac2bb909bd79f9759/pkg/auth/auth_test.go#L10 

Really silly of me to forget.

pkg/auth/auth.go
williammartin: > Any concerns about cli/go-gh not 1) normalizing the hostname as part of the boolean checks

Gah, I don't know. Some normalization does happen in `go-gh` so maybe? I just don't have a clear picture of when this happens and why it is important. It's probably best to match behaviour though even if we don't fully understand. https://github.com/cli/go-gh/blob/d88d88f917cd4e0bcaa582bac2bb909bd79f9759/pkg/auth/auth.go#L155-L164

> 2) extracting the tenant name from a host name like [cli/cli](https://github.com/cli/cli/blob/24481c5c2e49d3529a6621953b8878fe0514ff22/internal/ghinstance/host.go#L23-L60)?

Don't see any reason to have this behaviour, it's used for something to do with repo [remote URLs and `SSH`](https://github.com/cli/cli/blob/24481c5c2e49d3529a6621953b8878fe0514ff22/internal/ghrepo/repo.go#L97) (though I do wonder if the same [thing should be happening for gists, that's a separate issue.](https://github.com/cli/cli/blob/24481c5c2e49d3529a6621953b8878fe0514ff22/pkg/cmd/gist/clone/clone.go#L96)

pkg/auth/auth.go
```diff
@@ -160,5 +168,17 @@ func normalizeHostname(host string) string {
...
 		return localhost
 	}
+	if before, found := cutSuffix(hostname, "."+tenancyHost); found {
+		idx := strings.LastIndex(before, ".")
+		return fmt.Sprintf("%s.%s", before[idx+1:], tenancyHost)
+	}
 	return hostname
 }
+
+// Backport strings.CutSuffix from Go 1.20.
```
williammartin, line 182: Strictly speaking we shouldn't need this backport anymore but I'd rather change `cli/cli` and this at the same time, and I don't want to change `cli/cli` right now.

andyfeller:
> > Would we want to bring over the [isTenancy() tests](https://github.com/cli/cli/blob/24481c5c2e49d3529a6621953b8878fe0514ff22/internal/ghinstance/host_test.go#L52-L89) over too?
> 
> Not specifically but this reminded me that I definitely want to add tests to
> 
> https://github.com/cli/go-gh/blob/d88d88f917cd4e0bcaa582bac2bb909bd79f9759/pkg/auth/auth_test.go#L10
> 
> Really silly of me to forget.

I know, but I hate to say we have a test suite for `IsEnterprise` already:

https://github.com/cli/go-gh/blob/ee82f3a57020e59e18befbb7f39b5aae62cfeb41/pkg/auth/auth_test.go#L229-L258

You really sure?

andyfeller (approved):
Looks good all in all.

I'm going to defer to you if you really don't think bringing over the tests for `isTenancy` is necessary here.

williammartin:
Bleh. So firstly, I have to say that the existing tests are already too knowledgable about implementation details. That the public function `TokenForHost` is totally untested is an indictment on the current state of these tests.

image

`TokenForHost` is what users of the `go-gh` module rely on. That's the function that they expect to work. Testing the private functions that support `TokenForHost` instead not only reduces our ability to avoid regressions but actually congeals the existing implementation, increasing maintenance costs.

I'd added cases to the private `tokenForHost` test try and get as much of a black box test as possible within the current framework. At your prompting I've also added a test to communicate intent around how `isEnterprise` should handle tenants. I don't think there's much value in adding tests around `isTenant`.

> I know, but I hate to say we have a test suite for IsEnterprise already

Yeh and I don't think it's good. What I really want to do is rework all the tests to avoid them knowing about any private functions but I don't have the appetite to tackle that here.

---

All that said, I really appreciate you pushing on the tests angle 😊

PR #153, opened 2024-03-19
Attempt to use user configured dirs for caching

williammartin:
## Description

This change supports `XDG_CACHE_DIR` from the [xdg spec](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html).

A minor concern I have about moving this out of `/tmp` is that we might slowly fill up the disk. In practice I'm not sure that this is a big issue.

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -21,6 +21,7 @@ const (
 	xdgConfigHome = "XDG_CONFIG_HOME"
 	xdgDataHome   = "XDG_DATA_HOME"
 	xdgStateHome  = "XDG_STATE_HOME"
+	xdgCacheHome  = "XDG_CACHE_HOME"
 )
 
 var (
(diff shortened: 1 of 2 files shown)


pkg/config/config.go
```diff
@@ -287,9 +288,21 @@ func DataDir() string {
 	return path
 }
 
-// CacheDir returns the default path for gh cli cache.
+// Cache path precedence: XDG_CACHE_HOME, LocalAppData (windows only), HOME.
```
andyfeller: Is it worth adding the lowest precedent?

```suggestion
// Cache path precedence: XDG_CACHE_HOME, LocalAppData (windows only), HOME, legacy gh-cli-cache.
```

pkg/config/config.go
```diff
@@ -287,9 +288,21 @@ func DataDir() string {
...
+		return filepath.Join(b, "GitHub CLI")
+	} else if c, err := os.UserHomeDir(); err == nil {
+		return filepath.Join(c, ".cache", "gh")
+	} else {
+		// Note that this has a minor security issue because /tmp is world-writeable.
+		// As such, it is possible for other users on a shared system to overwrite cached data.
+		// The practical risk of this is low, but it's worth calling out as a risk.
+		// I've included this here for backwards compatibility but we should consider removing it.
+		return filepath.Join(os.TempDir(), "gh-cli-cache")
+	}
```
andyfeller, line 305: I don't think there is a problem, but I was wondering how `cli/cli` would handle this directory if it didn't exist because someone deleted it manually or used [`gh config clear-cache`](https://github.com/cli/cli/blob/trunk/pkg/cmd/config/clear-cache/clear_cache.go) but I realized that would be an issue regardless where the directory was.

andyfeller (approved):
Not going to be a problem with this review, but `cli/cli` has [some hardcoded references to the legacy cache location](https://github.com/search?q=repo%3Acli%2Fcli%20%22gh-cli-cache%22&type=code)

williammartin:
> Not going to be a problem with this review, but cli/cli has [some hardcoded references to the legacy cache location](https://github.com/search?q=repo%3Acli%2Fcli%20%22gh-cli-cache%22&type=code)

Excellent call out. Let's make sure we resolve that when we bump!

PR #154, opened 2024-03-20
Upgrade to go 1.22

williammartin:
Following the pattern from https://github.com/cli/go-gh/commit/a1e5851644f999a592dd2a59d20233617ebdf61c

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -18,16 +18,16 @@ jobs:
       - name: Checkout repository
         uses: actions/checkout@v4
 
+      - name: Setup go
+        uses: actions/setup-go@v5
+        with:
+          go-version-file: 'go.mod'
+
       - name: Initialize CodeQL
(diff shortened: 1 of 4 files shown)


.github/workflows/codeql-analysis.yml
```diff
@@ -24,5 +24,10 @@ jobs:
           languages: go
           queries: security-and-quality
 
+      - name: Setup go
```
williammartin: https://github.com/github/codeql/issues/15647#issuecomment-2003768106

williammartin:
I'm closing this because since `1.21`, this version is no longer advisory (https://go.dev/doc/modules/gomod-ref):

> The go directive sets the minimum version of Go required to use this module. Before Go 1.21, the directive was advisory only; now it is a mandatory requirement: Go toolchains refuse to use modules declaring newer Go versions.

It doesn't seem necessary to force our consumers to bump their go.mod versions.

PR #159, opened 2024-04-29
Use `lipgloss.Width` to calculate `DisplayWidth`

maaslalani:
Fixes https://github.com/cli/cli/issues/9018

Uses `lipgloss.Width` to calculate `DisplayWidth` which correctly aligns emojis.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -6,6 +6,7 @@ require (
 	github.com/AlecAivazis/survey/v2 v2.3.7
 	github.com/MakeNowJust/heredoc v1.0.0
 	github.com/charmbracelet/glamour v0.7.0
+	github.com/charmbracelet/lipgloss v0.10.1-0.20240413172830-d0be07ea6b9c
 	github.com/cli/browser v1.3.0
 	github.com/cli/safeexec v1.0.0
 	github.com/cli/shurcooL-graphql v0.0.4
@@ -17,7 +18,7 @@ require (
 	github.com/muesli/termenv v0.15.2
 	github.com/stretchr/testify v1.7.0
 	github.com/thlib/go-timezone-local v0.0.0-20210907160436-ef149e42d28e
-	golang.org/x/sys v0.18.0
+	golang.org/x/sys v0.19.0
 	golang.org/x/term v0.13.0
 	golang.org/x/text v0.13.0
 	gopkg.in/h2non/gock.v1 v1.1.2
@@ -28,6 +29,7 @@ require (
 	github.com/alecthomas/chroma/v2 v2.8.0 // indirect
 	github.com/aymanbagabas/go-osc52/v2 v2.0.1 // indirect
 	github.com/aymerick/douceur v0.2.0 // indirect
+	github.com/charmbracelet/x/exp/term v0.0.0-20240425164147-ba2a9512b05f // indirect
 	github.com/davecgh/go-spew v1.1.1 // indirect
 	github.com/dlclark/regexp2 v1.4.0 // indirect
 	github.com/gorilla/css v1.0.0 // indirect
--- a/pkg/text/text.go
+++ b/pkg/text/text.go
@@ -8,8 +8,8 @@ import (
 	"time"
 	"unicode"
 
-	"github.com/muesli/reflow/ansi"
 	"github.com/muesli/reflow/truncate"
+	"github.com/rivo/uniseg"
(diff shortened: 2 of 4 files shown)


maaslalani:
image

pkg/text/text.go
```diff
@@ -33,7 +33,7 @@ func Indent(s, indent string) string {
 
 // DisplayWidth calculates what the rendered width of string s will be.
 func DisplayWidth(s string) int {
-	return ansi.PrintableRuneWidth(s)
+	return uniseg.StringWidth(s)
```
williammartin: **Change Requested**

Can you add the emoji from https://github.com/cli/cli/issues/9018 into the `TestDisplayWidth` test cases?

williammartin (changes requested):

maaslalani:
@williammartin Had to switch to `lipgloss.Width` to account for ANSI sequences since `uniseg` doesn't strip ANSI before calculating width (one of the test cases included ANSI sequences), but `lipgloss` does then uses `uniseg.StringWidth` under the hood.

Added the multi-byte emoji test case as well!

go.mod
```diff
@@ -17,7 +18,7 @@ require (
 	github.com/muesli/termenv v0.15.2
 	github.com/stretchr/testify v1.7.0
 	github.com/thlib/go-timezone-local v0.0.0-20210907160436-ef149e42d28e
-	golang.org/x/sys v0.18.0
+	golang.org/x/sys v0.19.0
```
maaslalani, line 21: Let me know if you don't want to bump this.

williammartin:
Thanks. If we're expecting `DisplayWidth` to handle ANSI escape sequences then let's add a test for that as well.

I suppose for someone out there this could be a breaking change but I think it's edge case enough that I'd rather just accept it and then people could fix any hacky workarounds they might have had.

maaslalani:
Yep, there is already a test case for that! The tests were failing when I simply had `uniseg.StringWidth` because of the ANSI test case. Which is why I needed to swap out for lipgloss to pass all tests cases.

maaslalani:
Test case for ignoring ANSI sequences when calculating width is here: https://github.com/cli/go-gh/blob/trunk/pkg/text/text_test.go#L282-L287 (fails when _just_ using `uniseg.StringWidth`, passes when using `lipgloss.Width`)

williammartin:
Well, that's testing the public API of `PadRight` but it's just an implementation detail that `PadRight` uses `DisplayWidth`. Refactoring `PadRight` would then lose all our confidence that `DisplayWidth` handles ANSI codes. I want to both ensure that `DisplayWidth` enforces this behaviour, and I want it to be communicated to future maintainers in the tests because I had no idea that it was expected when you first opened the PR.

maaslalani:
Ah you're totally right, there is also a test case directly for color codes in `DisplayWidth` as well:

https://github.com/cli/go-gh/blob/5840c447a66255f2822dfc9be315be0f9ff1d6d2/pkg/text/text_test.go#L360-L363

Happy to add more test cases though if you want / prefer!

williammartin:
Haha sorry I should have also looked closer. I think I got red herring-ed by your link because I assumed if it existed already you would definitely have linked it 😅 

I can confirm that test fails just using `uniseg.StringWidth`:

```
--- FAIL: TestDisplayWidth (0.00s)
    --- FAIL: TestDisplayWidth/color_codes (0.00s)
        /Users/williammartin/workspace/go-gh/pkg/text/text_test.go:388: 
            	Error Trace:	text_test.go:388
            	Error:      	Not equal: 
            	            	expected: 3
            	            	actual  : 12
            	Test:       	TestDisplayWidth/color_codes
```

No need for extra tests, this looks great to me, once I pull it into `cli/cli` and play with it, cheers.

maaslalani:
> I think I got red herring-ed by your link because I assumed if it existed already you would definitely have linked it 😅

Haha all good! I kind of missed that test case the first time as well and definitely should have linked to that!

williammartin:
### Before

image

(wow such codepoints: https://ardislu.dev/biggest-unicode-grapheme-cluster)

### After

image

williammartin (approved):
I pulled this into `cli/cli` and it passed all the tests.

LGTM thanks a lot.

Abeabenut (commented):
https://github.com/cli/cli/releases/tag/v2.48.0

PR #162, opened 2024-05-19
fix: handle empty response body

### LINKED ISSUE #161: Rest client support for endpoint not returning a JSON
# Problem statement

While working on [`gh-not`](https://github.com/nobe4/gh-not) I started experimenting with Actors interacting with the GitHub API.

A very simple one would be to mark a notification as `read`, as per https://docs.github.com/en/rest/activity/notifications?apiVersion=2022-11-28#mark-a-thread-as-read.

Here's a simplified version of the code I wanted to use:

```go

func main() {
	client, err := api.DefaultRESTClient()
	if err != nil {
		panic(err)
	}

	url := "https://api.github.com/notifications/threads/[REDACTED]"

	resp := []interface{}{}
	err = client.Patch(url, nil, &resp)
	if err != nil {
		panic(err)
	}

}
```

And I always get: `panic: unexpected end of JSON input`

# Possible fix 1

The [documentation](https://docs.github.com/en/rest/activity/notifications?apiVersion=2022-11-28#mark-a-thread-as-read--status-codes) specifies status codes, but no response content; expecting one is thus useless in this context.

I wonder if https://github.com/cli/go-gh/blob/dbd982eba2a041d88d398cce01cb708c4c3503f7/pkg/api/rest_client.go#L103-L111 could be modified in the following way:

```diff
	b, err := io.ReadAll(resp.Body)
	if err != nil {
		return err
	}

+        if len(b) > 0{
        	err = json.Unmarshal(b, &response)
        	if err != nil {
        		return err
        	}
+        }
```

This would prevent breaking on invalid JSON when none is expected.

POC implemented in https://github.com/cli/go-gh/pull/162

# Possible fix 2

The [`205` status code spec](https://httpwg.org/specs/rfc9110.html#status.205) specifies:

> a server MUST NOT generate content in a 205 response

So maybe a better fix would be:

```diff
	if resp.StatusCode == http.StatusNoContent {
		return nil
	}

+	if resp.StatusCode == http.StatusResetContent {
+		return nil
+	}
```

Implementation PR: https://github.com/cli/go-gh/pull/163

# Current workaround

```go
func main() {
	client, err := api.DefaultRESTClient()
	if err != nil {
		panic(err)
	}

	url := "https://api.github.com/notifications/threads/[REDACTED]"

	resp := []interface{}{}
	err = client.Patch(url, nil, &resp)
	if err != nil && err.Error() != "unexpected end of JSON input" {
		panic(err)
	}
        // continue as planned
}
```

# Additional questions

- Is there another way to use `go-gh`'s REST client for Patch requests if the endpoint doesn't return a valid JSON?
- Would there be any benefits in returning the response code from those requests? It seems that implementing custom logic solely based on JSON breaks as soon as no JSON is returned

nobe4:
This PR is a showcase of solution 1 for https://github.com/cli/go-gh/issues/161.

It adds an empty body check before trying to unmarshall it.

### DIFF
--- a/pkg/api/rest_client.go
+++ b/pkg/api/rest_client.go
@@ -105,9 +105,11 @@ func (c *RESTClient) DoWithContext(ctx context.Context, method string, path stri
 		return err
 	}
 
-	err = json.Unmarshal(b, &response)
-	if err != nil {
-		return err
+	if len(b) > 0 {
+		err = json.Unmarshal(b, &response)
+		if err != nil {
+			return err
+		}
 	}
 
 	return nil
--- a/pkg/api/rest_client_test.go
+++ b/pkg/api/rest_client_test.go
@@ -303,6 +303,25 @@ func TestRESTClientPatch(t *testing.T) {
 	assert.True(t, gock.IsDone(), printPendingMocks(gock.Pending()))
 }
 
+// Covers https://docs.github.com/en/rest/activity/notifications?apiVersion=2022-11-28#mark-a-thread-as-read
+func TestRESTClientPatchNoResponseBody(t *testing.T) {
+	t.Cleanup(gock.Off)
+	gock.New("https://api.github.com").
+		Patch("/some/path/here").
+		BodyString(`{}`).
+		Reply(205).
+		BodyString("")
+	client, _ := NewRESTClient(ClientOptions{
+		Host:      "github.com",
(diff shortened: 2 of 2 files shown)


babakks:
Closing as this in favour of the method implemented in #163.

PR #163, opened 2024-05-19
fix: handle HTTP status 205 properly

### LINKED ISSUE #161: Rest client support for endpoint not returning a JSON

nobe4:
As discussed in https://github.com/cli/go-gh/issues/161, this might be one other fix for supporting empty bodies from the server.

The test is a POC, and should probably be refactored into something better.


joelazar:
Any updates on this? I think this is a clear fix and would love to see this in the lib, I have nightmares from 🙀
```go
	if err != nil && err.Error() != "unexpected end of JSON input" {
```

BagToad:
Hey @nobe4, @babakks and are looking at this issue/PR and agree with this approach. I'm wondering if there was something that was on your mind in particular about what might need to be polished up more? It seems like you're following convention here, and it might be good as-is.

I know that this was some time ago 🙏 so if this isn't relevant any longer, I understand. Looking to see if we couldn't land this in a `go-gh` release sometime soon!

nobe4:
Hi @BagToad :wave: 

I honestly don't recall what might have been improved :grimacing:
So if you think this PR is good as-is, feel free to move it forward :)

Thanks :pray:

babakks (approved):
Thanks for this, @nobe4! 🙏

PR #172, opened 2024-09-13
Use api subdomain for REST and GQL clients when host is tenant

williammartin:
## Description

When targeting the API of tenants we should be using the subdomain rather than the path prefix.

### Implementation Choices

Right now, I took the straightest path to implementing this, which included copying functions from the `auth` package. It was an intentional choice not to DRY this out because I want to derisk this work and then review it with confidence that the black box behaviour is correct.

### How To Test

The easiest way to black box validate these changes is to use the PR in `cli/cli` that pins to the [commit sha.](https://github.com/cli/cli/pull/9618)

Alternatively, replace the dependency in `cli/cli` like so:

```
go mod edit -replace github.com/cli/go-gh/v2=/path/to/go-gh
```

Then build and run the CLI with commands that use these clients such as:

```
GH_DEBUG=api GH_HOST=tenant.ghe.com ./bin/gh secret list --repo whatever/whatever # REST
GH_DEBUG=api GH_HOST=tenant.ghe.com ./bin/gh repo list # GQL
```

In both cases the HTTP requests should demonstrate the API subdomain being used instead of the path prefix.

### DIFF
--- a/pkg/api/http_client.go
+++ b/pkg/api/http_client.go
@@ -135,7 +135,14 @@ func isGarage(host string) bool {
 }
 
 func isEnterprise(host string) bool {
-	return host != github && host != localhost
+	return host != github && host != localhost && !isTenancy(host)
+}
+
+// tenancyHost is the domain name of a tenancy GitHub instance.
+const tenancyHost = "ghe.com"
+
+func isTenancy(host string) bool {
+	return strings.HasSuffix(host, "."+tenancyHost)
 }
 
 func normalizeHostname(hostname string) string {
@@ -145,6 +152,14 @@ func normalizeHostname(hostname string) string {
 	}
 	if strings.HasSuffix(hostname, "."+localhost) {
 		return localhost
+	}
+	// This has been copied over from the cli/cli NormalizeHostname function
+	// to ensure compatible behaviour but we don't fully understand when or
+	// why it would be useful here. We can't see what harm will come of
+	// duplicating the logic.
+	if before, found := strings.CutSuffix(hostname, "."+tenancyHost); found {
+		idx := strings.LastIndex(before, ".")
+		return fmt.Sprintf("%s.%s", before[idx+1:], tenancyHost)
 	}
 	return hostname
 }
--- a/pkg/api/graphql_client_test.go
+++ b/pkg/api/graphql_client_test.go
@@ -293,6 +293,11 @@ func TestGraphQLEndpoint(t *testing.T) {
 			host:         "enterprise.com",
 			wantEndpoint: "https://enterprise.com/api/graphql",
 		},
+		{
+			name:         "tenant",
+			host:         "tenant.ghe.com",
+			wantEndpoint: "https://api.tenant.ghe.com/graphql",
+		},
 	}
 
 	for _, tt := range tests {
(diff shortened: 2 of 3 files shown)


pkg/api/http_client.go
```diff
@@ -135,7 +135,14 @@ func isGarage(host string) bool {
...
 func isEnterprise(host string) bool {
-	return host != github && host != localhost
+	return host != github && host != localhost && !isTenancy(host)
+}
+
+// tenancyHost is the domain name of a tenancy GitHub instance.
+const tenancyHost = "ghe.com"
+
+func isTenancy(host string) bool {
+	return strings.HasSuffix(host, "."+tenancyHost)
```
andyfeller, line 145: **Are there any concerns about this being seemingly redundant to the logic in `auth`?**

I did a double take when I first saw this because I thought the enterprise check was already taking tenancy into account, when I realized I was thinking about:

https://github.com/cli/go-gh/blob/25db6b99518c88e03f71dbe9e58397c4cfb62caf/pkg/auth/auth.go#L152-L167

My concern is managing 2 sets of similar / overlapping functionality in 2 places and introducing drift.

pkg/api/http_client.go
```diff
@@ -146,6 +153,14 @@ func normalizeHostname(hostname string) string {
...
 		return localhost
 	}
+	// This has been copied over from the cli/cli NormalizeHostname function
+	// to ensure compatible behaviour but we don't fully understand when or
+	// why it would be useful here. We can't see what harm will come of
+	// duplicating the logic.
+	if before, found := strings.CutSuffix(hostname, "."+tenancyHost); found {
+		idx := strings.LastIndex(before, ".")
+		return fmt.Sprintf("%s.%s", before[idx+1:], tenancyHost)
+	}
```
andyfeller, line 163: Similar to my comment above, it seems we now have 3 copies of this logic across `cli/cli` and 2 places within `cli/go-gh` and that makes me a little concerned:

https://github.com/cli/go-gh/blob/25db6b99518c88e03f71dbe9e58397c4cfb62caf/pkg/auth/auth.go#L169-L186

andyfeller (commented):
I'm torn regarding some of the duplication of this logic within `cli/go-gh` as well as `cli/cli`.  My knee jerk would be `cli/go-gh` is the source of truth for this logic and `cli/cli` leveraging the logic for its needs, which I realize is outside of the scope of the original issue and likely would be deferred.

Beyond that, I wonder how we go about testing these changes in an extension other than `gh` to ensure they work as expected.  Perhaps this is a great opportunity to pair on testing this with `gh copilot`?

pkg/api/http_client.go
williammartin: See "implementation choices" section in PR description.

I think we should do work here or in a follow up to address this but I'm not going through tearing things up until we know this is works as expected and covers the right use cases.

pkg/api/http_client.go
williammartin: Same as my response above.

williammartin:
> Beyond that, I wonder how we go about testing these changes in an extension other than gh to ensure they work as expected. Perhaps this is a great opportunity to pair on testing this with gh copilot?

I'm not sure what the difference between `gh` and extensions are though, they both consume this module in the same way. However, it's probably worth doing anyway to derisk.

Best way would be the same way I suggested in the PR description for `gh`.

andyfeller (approved):
LGTM

PR #173, opened 2024-10-10
Replace duplicated IsEnterprise and IsTenancy functions in api pkg with auth pkg

jtmcg:
Related to github/cli#577

## Changes

In attempting to use `go-gh` in `cli/cli` for `IsEnterprise` and `IsTenancy`, I got bamboozled by the duplicate code, here. This removes the duplicated, unexported `isEnterprise` and `isTenancy` functions from the `api` package and replaces them with the exported `IsEnterprise` and `IsTenancy` packages from `auth`.

## Testing

1. Pull down branch
2. Run tests

Note, I've also tested these in cli/cli by running `go mod edit --replace=github.com/cli/go-gh=<localPath>/go-gh` and swapped these functions in. Everything there seems to be working as expected, too. I ran all the tests and did some `gh auth` testing, just to be sure


pkg/auth/auth.go
```diff
@@ -26,6 +26,8 @@ const (
 	hostsKey              = "hosts"
 	localhost             = "github.localhost"
 	oauthToken            = "oauth_token"
+	// TenancyHost is the domain name of a tenancy GitHub instance.
+	tenancyHost = "ghe.com"
```
andyfeller: Looking at this I asked myself _"Does VS Code really not try to fix the formatting on this one?"_ only to realize, no it doesn't.

Not the end of the world, but would rather the longer line with consistent formatting than shorter lines that are inconsistent. Not end of the world 🤷 

```suggestion
	tenancyHost           = "ghe.com" // TenancyHost is the domain name of a tenancy GitHub instance.
```

andyfeller (approved):
shipit-squirrel-thumbsup

williammartin (approved):
Seems reasonable to me. Maybe there was an original intention here to adhere to

> A little copying is better than a little dependency (https://go-proverbs.github.io/)

But even if it was, I think that this has been a source of annoyance a number of times, so let's just do it!

pkg/auth/auth.go
jtmcg: Seems reasonable to me as it also bothered me 😅

PR #178, opened 2025-01-20
Add useful template functions to jq filters

heaths:
Relates to #10262 but needs to be used in cli/cli if accepted.


heaths (commented):
@0xdevalias if acceptable to the CLI devs, this could be the basis for using *some* `--template` functions in `--jq` filters. Most of them don't make sense or have equivalents already in jq, though: https://jqlang.github.io/jq/manual/

If acceptable, I can document this in the CLI's README and help content. I think even what's there could use some sprucing up anyway e.g., to link acceptable formats to https://pkg.go.dev/time#Layout.

0xdevalias:
> if acceptable to the CLI devs, this could be the basis for using _some_ `--template` functions in `--jq` filters

> If acceptable, I can document this in the CLI's README and help content

@heaths Sounds good to me! Appreciate your efforts! 🖤

0xdevalias:
> Relates to #10262 but needs to be used in cli/cli if accepted.

Proper link to issue (since it's in a different repo):

- https://github.com/cli/cli/issues/10262

---

@andyfeller Curious if you (or someone more appropriate on your team perhaps) had some bandwidth to give a /2c on this PR in general; particularly around if/what concerns/blockers there would be to landing it (aside from the obvious fact that it needs a rebase; but if there are larger concerns with the feature/approach then that sort of 'polish'/'upkeep' would be irrelevant)

I know there is this issue for some of the more tangentially related issues connected to this; but looking at the changes here, I don't think this PR would really fall under the scope of the concerns raised in it:

- https://github.com/cli/cli/issues/10558

fc806063 (commented):
[heaths:issue10262](url)

heaths:
Any interest in this? I can rebase it and then its companion in cli/cli if someone will commit to looking at it shortly thereafter.

PR #180, opened 2025-02-07
Add accessibility package and isEnabled function

jtmcg:
## Changes

- Moved duplicated `stubConfig()` function to util
- Added `accessibility` package and `IsEnabled()` function

## Testing

This isn't wired up to anything yet, so unless its wired up to its own binary it is just covered by the unit tests. I think this is fine given we're merging into a feature branch right now.

### DIFF
--- a/internal/testutils/config_stub.go
+++ b/internal/testutils/config_stub.go
@@ -1,0 +1,21 @@
+package testutils
+
+import (
+	"testing"
+
+	"github.com/cli/go-gh/v2/pkg/config"
+)
+
+// StubConfig replaces the config.Read function with a function that returns a config object
+// created from the given config string. It also sets up a cleanup function that restores the
+// original config.Read function.
+func StubConfig(t *testing.T, cfgStr string) {
+	t.Helper()
+	old := config.Read
+	config.Read = func(_ *config.Config) (*config.Config, error) {
+		return config.ReadFromString(cfgStr), nil
+	}
+	t.Cleanup(func() {
+		config.Read = old
+	})
+}
--- a/pkg/accessibility/accessibility.go
+++ b/pkg/accessibility/accessibility.go
@@ -6,7 +6,11 @@ import (
 	"github.com/cli/go-gh/v2/pkg/config"
 )
 
-const ACCESSIBILITY_ENV = "ACCESSIBILITY"
+// ACCESSIBILITY_ENV is the name of the environment variable that can be used to enable
+// accessibility features. If the value is empty, "0", or "false", the accessibility
+// features are disabled. Any other value enables the accessibility features. Note that
+// this environment variable supercedes the configuration file's accessible setting.
+const ACCESSIBILITY_ENV = "GH_ACCESSIBILE"
 
 // IsEnabled returns true if accessibility features are enabled via the ACCESSIBILITY
 // environment variable or the configuration file.
--- a/pkg/accessibility/accessibility.go
+++ b/pkg/accessibility/accessibility.go
@@ -10,7 +10,7 @@ import (
 // accessibility features. If the value is empty, "0", or "false", the accessibility
 // features are disabled. Any other value enables the accessibility features. Note that
 // this environment variable supercedes the configuration file's accessible setting.
-const ACCESSIBILITY_ENV = "GH_ACCESSIBILE"
+const ACCESSIBILITY_ENV = "ACCESSIBILE"
 
 // IsEnabled returns true if accessibility features are enabled via the ACCESSIBILITY
 // environment variable or the configuration file.
--- a/pkg/accessibility/accessibility.go
+++ b/pkg/accessibility/accessibility.go
@@ -1,0 +1,29 @@
+package accessibility
+
+import (
+	"os"
+
+	"github.com/cli/go-gh/v2/pkg/config"
+)
+
+const ACCESSIBILITY_ENV = "ACCESSIBILITY"
+
+// IsEnabled returns true if accessibility features are enabled via the ACCESSIBILITY
+// environment variable or the configuration file.
+func IsEnabled() bool {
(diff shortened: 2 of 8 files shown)


pkg/accessibility/accessibility.go
```diff
@@ -0,0 +1,29 @@
+package accessibility
+
+import (
+	"os"
+
+	"github.com/cli/go-gh/v2/pkg/config"
+)
+
+const ACCESSIBILITY_ENV = "ACCESSIBILITY"
```
andyfeller: Are there any concerns aligning with the `ACCESSIBLE` env var used by [`charmbracelet/huh`](https://github.com/charmbracelet/huh#accessibility) for accessible prompting?

```suggestion
const ACCESSIBILITY_ENV = "ACCESSIBLE"
```

pkg/accessibility/accessibility.go
```diff
@@ -0,0 +1,29 @@
...
+
+// IsEnabled returns true if accessibility features are enabled via the ACCESSIBILITY
+// environment variable or the configuration file.
```
andyfeller, line 16: Love the comments ❤️  Rather than including the env var, do you know if Go has a way of linking comments to types/constants/functions/etc?

pkg/accessibility/accessibility_test.go
```diff
@@ -0,0 +1,91 @@
...
+		},
+		{
+			name:        "When the accessibility configuration is unset and the ACCESSIBLE env var returns '0', it should return false",
+			envVarValue: "0",
+			cfgStr:      "",
+			wantOut:     false,
+		},
+		{
+			name:        "When the accessibility configuration is unset and the ACCESSIBLE env var returns 'false', it should return false",
+			envVarValue: "0",
```
andyfeller: Should this be `"false"`?

```suggestion
			name:        "When the accessibility configuration is unset and the ACCESSIBLE env var returns 'false', it should return false",
			envVarValue: "false",
```

pkg/accessibility/accessibility_test.go
```diff
@@ -0,0 +1,91 @@
...
+		},
+		{
+			name:        "When the accessibility configuration is set to disabled and the env var is set to something truthy (not '0' or 'false'), it should return true",
+			envVarValue: "true",
+			cfgStr:      accessibilityDisabledConfig(),
+			wantOut:     true,
+		},
+		{
+			name:        "When the accessibility configuration is set to enabled and the env var is set to '0', it should return false",
+			envVarValue: "false",
```
andyfeller: Should this be `"0"`?

```suggestion
			name:        "When the accessibility configuration is set to enabled and the env var is set to '0', it should return false",
			envVarValue: "0",
```

internal/testutils/config_stub.go
```diff
@@ -0,0 +1,21 @@
...
+	"testing"
+
+	"github.com/cli/go-gh/v2/pkg/config"
+)
+
+// StubConfig replaces the config.Read function with a function that returns a config object
+// created from the given config string. It also sets up a cleanup function that restores the
+// original config.Read function.
+func StubConfig(t *testing.T, cfgStr string) {
+	t.Helper()
```
andyfeller, line 13: awyiss

pkg/repository/repository_test.go
```diff
@@ -189,14 +189,3 @@ func TestParseWithHost(t *testing.T) {
...
-func stubConfig(t *testing.T, cfgStr string) {
-	t.Helper()
-	old := config.Read
-	config.Read = func(_ *config.Config) (*config.Config, error) {
-		return config.ReadFromString(cfgStr), nil
-	}
-	t.Cleanup(func() {
-		config.Read = old
-	})
-}
```
andyfeller, line 202: Really appreciate DRYing this up

andyfeller (changes requested):
Color me impressed with how quick you got this up 😮 Handful of minor things I suspect are quick to knock out.

pkg/accessibility/accessibility.go
jtmcg: Ah, good call. We should probably suffix this with `GH_` as well 👍

pkg/accessibility/accessibility.go
jtmcg: Good question... I _think_ that anything exported with a comment block will link the comment. Let me check on that....

pkg/accessibility/accessibility_test.go
jtmcg: Haha yup! I mixed my `envVarValues` between the two 😅 What's important, though, is that we're checking `0` and `false` in these different tests

pkg/accessibility/accessibility.go
jtmcg: And confirmed! I've added a comment block to the top of the `GH_ACCESSIBLE` env var so that it can be referenced by the editor 👍

andyfeller (approved):
👨‍🍳 💋

pkg/accessibility/accessibility.go
```diff
@@ -0,0 +1,33 @@
...
+	"os"
+
+	"github.com/cli/go-gh/v2/pkg/config"
+)
+
+// ACCESSIBILITY_ENV is the name of the environment variable that can be used to enable
+// accessibility features. If the value is empty, "0", or "false", the accessibility
+// features are disabled. Any other value enables the accessibility features. Note that
+// this environment variable supercedes the configuration file's accessible setting.
+const ACCESSIBILITY_ENV = "ACCESSIBILE"
```
williammartin, line 13: This is spelt incorrectly. "ACCESSIBILE" -> "ACCESSIBLE"

andyfeller: Fixed in https://github.com/cli/go-gh/pull/182/files#r1954881696

PR #182, opened 2025-02-12
Add accessible themes

jtmcg:
## Changes

- Add accessibility file to markdown package:
  - Adds ansi color code enum
  - Adds accessible glamour themes
- Add check for accessibility to `WithTheme`
- Add tests to cover rendering outcomes

## Notes

The link colors weren't rendering correctly with the [provided markdown colors](https://github.com/github/cli/issues/784#issuecomment-2644214586). In particular, it seemed that the link color was overriding the link text color. As such, I've changed the link stylings so that the text renders in either the bright blue or bright cyan.

## Testing

1. Pull down this branch
2. In the `cli/cli` repo, add a `replace` directive pointing at your local version of `go-gh`: `go mod edit -replace=github.com/cli/go-gh/v2={YOUR_LOCAL_PATH}/go-gh`
3. Build the binary
4. Run `ACCESSIBLE=true ./bin/gh repo view`

### Dark Theme
<= current | accessible =>
Screenshot 2025-02-12 at 3 04 09 PM (2)

### Light Theme
<= current | accessible =>
Screenshot 2025-02-12 at 3 06 22 PM (2)


pkg/markdown/markdown_test.go
```diff
@@ -1,38 +1,103 @@
...
 	"github.com/stretchr/testify/assert"
 )
 
+// glamour theme colors found at https://github.com/charmbracelet/glamour/tree/master/styles
+const (
+	glamourLightH2Color     = "27"
+	glamourDarkH2Color      = "39"
+	customH2Color           = "61"
+	magentaEscapeCode       = "35"
+	brightMagentaEscapeCode = "95"
```
andyfeller: Is it important to comment or modify the constant names to reflect which color depth these codes relate to?

- `magentaEscapeCode` and `brightMagentaEscapeCode` in this case are both [3-bit / 4-bit codes](https://en.wikipedia.org/wiki/ANSI_escape_code#3-bit_and_4-bit)
- `glamourLightH2Color`, `glamourDarkH2Color`, and `customH2Color` are [8-bit codes](https://en.wikipedia.org/wiki/ANSI_escape_code#8-bit)

pkg/markdown/markdown_test.go
```diff
@@ -1,38 +1,103 @@
...
+			name:    "notty style",
+			text:    "## h2",
+			theme:   "none",
+			wantOut: "## h2", // no tty should maintain the original text
 		},
 		{
-			name:  "notty style",
-			text:  "some text",
-			theme: "none",
+			name:        "when the style env var is set, we override the theme whith that style",
```
andyfeller:
```suggestion
			name:        "when the style env var is set, we override the theme with that style",
```

pkg/markdown/markdown_test.go
```diff
@@ -1,38 +1,103 @@
...
+		theme            string
+		styleEnvVar      string
+		accessibleEnvVar string
+		wantOut          string
 	}{
 		{
-			name:  "light style",
-			text:  "some text",
-			theme: "light",
+			name:    "light style",
```
andyfeller: Would test names like the following be easier for others to follow?

```
			name:    "default glamour light style renders header in 8-bit dodgerblue",
```

pkg/markdown/markdown_test.go
```diff
@@ -1,38 +1,103 @@
...
+			text:    "## h2",
+			theme:   "light",
+			wantOut: fmt.Sprintf("\x1b[0m\x1b[38;5;%s;1mh2", glamourLightH2Color),
+		},
+		{
+			name:    "dark style",
```
andyfeller:
```
			name:    "default glamour dark style renders header in 8-bit deepskyblue",
```

pkg/markdown/markdown_test.go
```diff
@@ -1,38 +1,103 @@
...
+			text:    "## h2",
+			theme:   "dark",
+			wantOut: fmt.Sprintf("\x1b[0m\x1b[38;5;%s;1mh2", glamourDarkH2Color),
 		},
 		{
-			name:  "dark style",
-			text:  "some text",
-			theme: "dark",
+			name:    "notty style",
```
andyfeller:
```
			name:    "default glamour notty style renders header without color",
```

pkg/markdown/markdown_test.go
```diff
@@ -1,38 +1,103 @@
...
 		t.Run(tt.name, func(t *testing.T) {
-			_, err := Render(tt.text, WithTheme(tt.theme))
+			t.Setenv(accessibility.ACCESSIBILITY_ENV, tt.accessibleEnvVar)
+
+			if tt.styleEnvVar != "" {
+				tmpDir := t.TempDir()
+				path := filepath.Join(tmpDir, fmt.Sprintf("%s.json", tt.styleEnvVar))
+				os.WriteFile(path, []byte(customGlamourStyle(t)), 0644)
+				t.Setenv("GLAMOUR_STYLE", path)
+			}
```
andyfeller: Love how clean this is

pkg/markdown/accessibility.go
```diff
@@ -0,0 +1,129 @@
...
+		"brightBlack",
+		"brightRed",
+		"brightGreen",
+		"brightYellow",
+		"brightBlue",
+		"brightMagenta",
+		"brightCyan",
+		"brightWhite",
+	}[a]
+}
```
andyfeller: Is there a benefit to this approach over a constructor where both the code and name are provided together?  Or perhaps a limitation in the constructor approach that makes this necessary?

```go
type ANSIColor struct {
  Code *string
  Name string
}

const (
  Black = &ANSIColor{ ... }
  ...
)
```

pkg/markdown/accessibility.go
```diff
@@ -0,0 +1,129 @@
...
+	cfg.Code.Color = Red.ValuePtr()
+
+	// Image colors
+	cfg.Image.Color = BrightMagenta.ValuePtr()
+
+	// Horizontal rule colors
+	cfg.HorizontalRule.Color = White.ValuePtr()
+
+	return cfg
+}
```
andyfeller: No right or wrong answer, but I take it you think we should selectively override the colors versus duplicating the whole style and maintaining it separately?

andyfeller (commented):
@jtmcg : Really like what I'm seeing especially the tests!  Nothing majorly blocking, some curious questions for your thoughts on various topics.

pkg/markdown/markdown_test.go
jtmcg: 🤔 I'm not sure they're different for terminal escape codes, right? Like, the 4-bit codes are values 30-37 & 90-97, and they're the same number for 8-bit color sequences but are encoded differently as escape sequences. Maybe what would be better is extracting the whole escape sequence and renaming? So something like:

```suggestion
	glamourLightH2_8bitColorEscSeq     = "\x1b[38;5;27;"
	glamourDarkH2_8bitColorEscSeq      = "\x1b[38;5;39;"
	customH2_8bitColorEscSeq           = "\x1b[38;5;61;"
	magenta_4bitColorEscSeq       = "\x1b[35"
	brightMagenta_4bitColorEscSeq = "\x1b[95"
```

`#namesarehard`

pkg/markdown/accessibility.go
jtmcg: That is an excellent question... This was screaming `enum` to me when I implemented it, but I was a bit unhappy with the way it turned out.

I don't think I'm actually using `AnsiColor.Name()` anywhere anymore, though, so maybe its moot and I can just get away with the enum and `.String()` and `.StrPtr()` methods. 

However, maybe there's a case to be made to take this one step further and add the expected escape sequences to the struct, so something like:

```golang
type AnsiColor struct {
    Code *string
    Name string
    EscSeq string
}
```

But to answer your question, I think the enum approach is _slightly_ more efficient, as we don't have to save all the ansi code structures in memory - just the integers and the methods on the type. The pointers will only get instantiated if we call the `AccessibilityStyleConfig()` switch, and it will instantiate only the pointers it needs. That said, it's a tiny optimization that probably inconsequential here, and the breadcrumbing from the proposed struct above may improve grokability enough to justify the tradeoff.

pkg/markdown/accessibility.go
jtmcg: The side-effects of overwriting the glamour styles were giving me pause... Mostly, I think this makes following the logic of what's happening for the accessibility experience easier to follow. After we chatted the other day about "how will we know if we're light or dark?" and the answer being "let glamour decide" I thought it more helpful to put the accessible things where that decision happens and be explicit about how its happening. 

Compare my approach:

https://github.com/cli/go-gh/blob/ac8e4551e299c525e47b261544c8b82bb85fc2a4/pkg/markdown/markdown.go#L35-L50

To yours

https://github.com/cli/go-gh/blob/0c30d0f6193cdb5c27a0b5716e7416a8e77c47d0/pkg/markdown/markdown.go#L54-L68

I find the logic of "this is how we are choosing to use the accessible light/dark theme" much easier to understand in the former vs the latter.

What are your thoughts?

pkg/markdown/accessibility.go
jtmcg: I decided to keep the enum and remove the unnecessary `.Name()` method. This did a lot to clean this up, I think

pkg/accessibility/accessibility.go
```diff
@@ -10,7 +10,7 @@ import (
 // accessibility features. If the value is empty, "0", or "false", the accessibility
 // features are disabled. Any other value enables the accessibility features. Note that
 // this environment variable supercedes the configuration file's accessible setting.
-const ACCESSIBILITY_ENV = "ACCESSIBILE"
+const ACCESSIBILITY_ENV = "ACCESSIBLE"
```
jtmcg, line 13: 🤦

pkg/markdown/accessibility.go
andyfeller: > The side-effects of overwriting the glamour styles were giving me pause... Mostly, I think this makes following the logic of what's happening for the accessibility experience easier to follow.

Sorry, let me clarify.

I understand why these changes only override the styles if `WithTheme()` is called as someone using `markdown.Render()` without providing the `TermRenderer` will not get any color.

My question was primarily focused on what value we'd get by statically duplicating [`DarkStyleConfig`](https://github.com/charmbracelet/glamour/blob/30f279f0e4288419e4d6e5e7c3a30f2abe3d095c/styles/styles.go#L131-L339) and [`LightStyleConfig`](https://github.com/charmbracelet/glamour/blob/30f279f0e4288419e4d6e5e7c3a30f2abe3d095c/styles/styles.go#L342-L548) variables in `cli/go-gh` over run-time copying and overriding specific fields.

Not something that needs to be done; just curious for your thoughts.  I can imagine this might make sense in the future as we increase the scope of customizing the styles including code block rendering, modifiers, and/or changing the resulting format of certain elements.

pkg/markdown/markdown_test.go
andyfeller: > and they're the same number for 8-bit color sequences but are encoded differently as escape sequences

I'm not sure I understand as [wiki](https://en.wikipedia.org/wiki/ANSI_escape_code#8-bit) shows that 8-bit color codes for ANSI colors are not the same:

```
ESC[38;5;⟨n⟩m Select foreground color      where n is a number from the table below
ESC[48;5;⟨n⟩m Select background color
  0-  7:  standard colors (as in ESC [ 30–37 m)
  8- 15:  high intensity colors (as in ESC [ 90–97 m)
 16-231:  6 × 6 × 6 cube (216 colors): 16 + 36 × r + 6 × g + b (0 ≤ r, g, b ≤ 5)
232-255:  grayscale from dark to light in 24 steps
```

Snippet from previous color contrast testing script displaying 8-bit ANSI colors where `(n)` goes from 0-15:

```shell
echo -e "   8 | Black               | ESC[48;5;0m   [\033[48;5;0m    \033[m] | ESC[38;5;0m   \033[38;5;0m andyfeller \033[m"
echo -e "   8 | Red                 | ESC[48;5;1m   [\033[48;5;1m    \033[m] | ESC[38;5;1m   \033[38;5;1m andyfeller \033[m"
echo -e "   8 | Green               | ESC[48;5;2m   [\033[48;5;2m    \033[m] | ESC[38;5;2m   \033[38;5;2m andyfeller \033[m"
echo -e "   8 | Yellow              | ESC[48;5;3m   [\033[48;5;3m    \033[m] | ESC[38;5;3m   \033[38;5;3m andyfeller \033[m"
echo -e "   8 | Blue                | ESC[48;5;4m   [\033[48;5;4m    \033[m] | ESC[38;5;4m   \033[38;5;4m andyfeller \033[m"
echo -e "   8 | Magenta             | ESC[48;5;5m   [\033[48;5;5m    \033[m] | ESC[38;5;5m   \033[38;5;5m andyfeller \033[m"
echo -e "   8 | Cyan                | ESC[48;5;6m   [\033[48;5;6m    \033[m] | ESC[38;5;6m   \033[38;5;6m andyfeller \033[m"
echo -e "   8 | White               | ESC[48;5;7m   [\033[48;5;7m    \033[m] | ESC[38;5;7m   \033[38;5;7m andyfeller \033[m"
echo -e "   8 | Bright Black        | ESC[48;5;8m   [\033[48;5;8m    \033[m] | ESC[38;5;8m   \033[38;5;8m andyfeller \033[m"
echo -e "   8 | Bright Red          | ESC[48;5;9m   [\033[48;5;9m    \033[m] | ESC[38;5;9m   \033[38;5;9m andyfeller \033[m"
echo -e "   8 | Bright Green        | ESC[48;5;10m  [\033[48;5;10m    \033[m] | ESC[38;5;10m  \033[38;5;10m andyfeller \033[m"
echo -e "   8 | Bright Yellow       | ESC[48;5;11m  [\033[48;5;11m    \033[m] | ESC[38;5;11m  \033[38;5;11m andyfeller \033[m"
echo -e "   8 | Bright Blue         | ESC[48;5;12m  [\033[48;5;12m    \033[m] | ESC[38;5;12m  \033[38;5;12m andyfeller \033[m"
echo -e "   8 | Bright Magenta      | ESC[48;5;13m  [\033[48;5;13m    \033[m] | ESC[38;5;13m  \033[38;5;13m andyfeller \033[m"
echo -e "   8 | Bright Cyan         | ESC[48;5;14m  [\033[48;5;14m    \033[m] | ESC[38;5;14m  \033[38;5;14m andyfeller \033[m"
echo -e "   8 | Bright White        | ESC[48;5;15m  [\033[48;5;15m    \033[m] | ESC[38;5;15m  \033[38;5;15m andyfeller \033[m"
```

All that being said, let me catch up on the actual changes as I suspect I understand what you're getting at.

pkg/markdown/accessibility_test.go
```diff
@@ -0,0 +1,118 @@
...
+			c:    BrightGreen,
+			want: strPtr("10"),
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			assert.Equal(t, tt.want, tt.c.StrPtr())
+		})
+	}
+}
```
andyfeller, line 57: I'll admit this touches on one of those rough edges of Go testing: pointer testing.

What level of equivalence is `assert.Equal` checking?  same pointer?  both are pointers to the same type?  both are pointers whose values are equal?

The other question in this specific scenario is whether an arbitrary string pointer not created by `strPtr` should really be tested against.  Otherwise, this test feels like we're making sure a expected output matches the result of some internal function which is also under test itself.

pkg/markdown/accessibility_test.go
```diff
@@ -0,0 +1,118 @@
...
+	assert.Equal(t, BrightWhite.StrPtr(), cfg.H1.StylePrimitive.Color)
+	assert.Equal(t, BrightBlue.StrPtr(), cfg.H1.StylePrimitive.BackgroundColor)
+	assert.Equal(t, BrightWhite.StrPtr(), cfg.Code.BackgroundColor)
+	assert.Equal(t, Red.StrPtr(), cfg.Code.Color)
+	assert.Equal(t, BrightMagenta.StrPtr(), cfg.Image.Color)
+	assert.Equal(t, White.StrPtr(), cfg.HorizontalRule.Color)
+
+	// Test that we haven't changed the original style
+	assert.Equal(t, styles.LightStyleConfig.H2, cfg.H2)
+}
```
andyfeller: What are your thoughts if we are capable of building a more comprehensive test that can traverse [`ansi.StyleConfig` type](https://github.com/charmbracelet/glamour/blob/30f279f0e4288419e4d6e5e7c3a30f2abe3d095c/ansi/style.go#L98-L139) for any colors we did not override to ensure they are customizable?

_I realize that will be significant enough work to be a separate issue and PR, but want to gather your thoughts while we're here._

pkg/accessibility/accessibility.go
```diff
@@ -10,7 +10,7 @@ import (
...
 
 // IsEnabled returns true if accessibility features are enabled via the ACCESSIBILITY
```
andyfeller:
```suggestion
// IsEnabled returns true if accessibility features are enabled via the ACCESSIBLE
```

andyfeller (changes requested):
Only 1 minor change to fix and let's ship it!

I'm leaving comments regarding a few errant thoughts on other aspects but nothing that's blocking for now.  I look forward to seeing how our testing around style customization evolves.

pkg/markdown/accessibility_test.go
jtmcg: According to the [assert package docs](https://pkg.go.dev/github.com/stretchr/testify@v1.7.0/assert#Equal):

> Pointer variable equality is determined based on the equality of the referenced values (as opposed to the memory addresses). Function equality cannot be determined and will always fail.

So this is checking that the pointer value references the same string values, which is what we want.

pkg/markdown/accessibility_test.go
jtmcg: While it would be an interesting engineering problem, I don't think its one we should solve until we have a reason to, i.e. we screw something up and this suggestion would keep us from doing it again.

pkg/markdown/accessibility.go
jtmcg: Oh, I understand, like why not "copy/paste" the [darkStyleConfig](https://github.com/charmbracelet/glamour/blob/30f279f0e4288419e4d6e5e7c3a30f2abe3d095c/styles/styles.go#L132-L339) from glamour into `go-gh`, rename for accessibility, and use that instead of creating this at run-time?

There's two benefits to this, I think:
1. This makes it really clear how we're deviating from Glamour's styles for accessibility
2. Given we are talking about contributing back to Glamour for this, the less we have to maintain here the better.

andyfeller (approved):

PR #183, opened 2025-02-19
Ensure codeblocks use accessible colors

andyfeller:
This commit leverages `cli` fork of `charmbracelet/glamour` with enhancement for configuring `chroma` formatter.

It includes simple tests that enabling `gh` accessible features causes codeblocks to be downsampled to base 16 ANSI colors.


andyfeller:
Looking back at the changes in the PR versus https://github.com/cli/cli/pull/9821, I'm confirming the need to remove the `replace` and change all of the `import` statements to use our fork until we can get the changes accepted upstream.

pkg/markdown/markdown_test.go
```diff
@@ -74,6 +75,48 @@ func Test_Render(t *testing.T) {
...
+		{
+			name: "when the accessible env var is set and the dark theme is selected, the codeblock renders using 4-bit colors",
+			text: heredoc.Docf(`
+				%[1]s%[1]s%[1]sgo
+				fmt.Println("Hello, world!")
+				%[1]s%[1]s%[1]s
+			`, "`"),
+			theme:            "dark",
+			accessibleEnvVar: "true",
+			wantOut:          "\x1b[0m\x1b[30mfmt\x1b[0m\x1b[33m.\x1b[0m\x1b[36mPrintln\x1b[0m\x1b[33m(\x1b[0m\x1b[90m\"Hello, world!\"\x1b[0m\x1b[33m)\x1b[0m",
```
jtmcg: I find it really hard to tell what's different between this blob of output versus the one on line 96. That's why I added the human-grokkable format in the tests above these. I also find it a bit jarring that the output formats are different for these tests and would find value in their consistency. Perhaps codeblocks are different enough to justify it, but there may be value in pulling out a human-grokkable abstraction in these `wantOut`s

jtmcg (commented):
Could you link the fork changes, @andyfeller? I'd find it helpful so we understand what this is introducing from the fork.

pkg/markdown/markdown.go
```diff
@@ -39,7 +39,20 @@ func WithTheme(theme string) glamour.TermRendererOption {
...
+
+				return func(tr *glamour.TermRenderer) error {
+					if err := stylesOption(tr); err != nil {
+						return err
+					}
+					if err := chromaOption(tr); err != nil {
+						return err
+					}
+					return nil
+				}
```
jtmcg: Clever, but I really don't like how this reads... I find it pretty confusing given this is returning a `glamour.TermRendererOption`. I guess it's a function of this call signature, but that requires implementation level knowledge of glamour to understand this code.

I think I might prefer to abstract this closure out to something separate, so we end up with something like

```suggestion
				return AccessibleStyles(theme)

// in the accessibility dir

func AccessibleStyles(theme string) glamour.TermRendererOption {
    // Applying multiple glamour.TermRendererOption here requires a wrapper that applies each
    // within glamour.NewTermRenderer() in Render() below.
    stylesOption := glamour.WithStyles(AccessibleStyleConfig(theme))
    chromaOption := glamour.WithChromaFormatter("terminal16")
    
    return func(tr *glamour.TermRenderer) error {
	if err := stylesOption(tr); err != nil {
		return err
	}
	if err := chromaOption(tr); err != nil {
		return err
	}
	return nil
    }
}
```

That way, we're separating the logic of "this is how you build the accessible theme using our implementation" from "select the right theme to use"

pkg/markdown/markdown_test.go
andyfeller: Completely get that 💯 as that was a consideration when I was writing the tests I choose to defer to seeing if someone would comment on. 👍 

I can play around to see what this would look like for a more grokkable format but I suspect the extent of control characters is pretty significant.  I don't know how much more grokkable it will be trying to reuse a bunch of placeholders but we can try.

pkg/markdown/markdown.go
andyfeller: > Clever, but I really don't like how this reads... I find it pretty confusing given this is returning a `glamour.TermRendererOption`. I guess it's a function of this call signature, but that requires implementation level knowledge of glamour to understand this code.
> 
> I think I might prefer to abstract this closure out to something separate, so we end up with something like
> 
> That way, we're separating the logic of "this is how you build the accessible theme using our implementation" from "select the right theme to use"

I was definitely conflicted in the implementation because of where / how we were applying the accessible theme in `WithTheme` versus in `Render` where this would allow us to have separate `TermRendererOption` that would be cleaner.

That said, I think at some level the reader digging into this has to understand `glamour` as long as its the dependency we use here.

Do you think further abstracting this could create a different type of confusion for reasoning about this behavior?

andyfeller:
Going to close this PR for the moment because there is a bit more work to do with changing this to use our fork of `glamour` with `import` paths.

Will continue responding to comments in the mean time. 👍

pkg/markdown/markdown_test.go
andyfeller: Here's an example of what this would look like if we used the same approach of constants:

```go
	reset                                            = "\x1b[0m"
	glamourLightCodeBlock_8bitRaisinBlackColorSeq    = "\x1b[38;5;235m"
	glamourLightCodeBlock_8bitVividTangerineColorSeq = "\x1b[38;5;210m"
	glamourLightCodeBlock_8bitJadeColorSeq           = "\x1b[38;5;35m"
	glamourLightCodeBlock_8bitCopperRoseColorSeq     = "\x1b[38;5;95m"
```

```go
		{
			name: "when the light theme is selected, the codeblock renders using 8-bit colors",
			text: heredoc.Docf(`
				%[1]s%[1]s%[1]sgo
				fmt.Println("Hello, world!")
				%[1]s%[1]s%[1]s
			`, "`"),
			theme: "light",
			// wantOut: "\x1b[0m\x1b[38;5;235mfmt\x1b[0m\x1b[38;5;210m.\x1b[0m\x1b[38;5;35mPrintln\x1b[0m\x1b[38;5;210m(\x1b[0m\x1b[38;5;95m\"Hello, world!\"\x1b[0m\x1b[38;5;210m)\x1b[0m",
			wantOut: fmt.Sprintf("%[1]s%[2]sfmt%[1]s%[3]s.%[1]s%[2]sPrintln%[1]s%[3]s(%[1]s%[4]s\"Hello, world!\"%[1]s%[3]s)%[1]s", reset, glamourLightCodeBlock_8bitRaisinBlackColorSeq, glamourLightCodeBlock_8bitVividTangerineColorSeq, glamourLightCodeBlock_8bitCopperRoseColorSeq),
		},
```

andyfeller: I also think there is something weird going on with the test as I originally implemented the `dark` tests and it used different escape codes, however I think there is some flakiness with `chroma` between tests that it isn't using the appropriate theme.

Before introducing `light` test case, this was the result of `dark` without and with accessible colors:

```go
wantOut: "\x1b[0m\x1b[38;5;251mfmt\x1b[0m\x1b[38;5;187m.\x1b[0m\x1b[38;5;42mPrintln\x1b[0m\x1b[38;5;187m(\x1b[0m\x1b[38;5;173m\"Hello, world!\"\x1b[0m\x1b[38;5;187m)\x1b[0m",
```

```go
wantOut: "\x1b[0m\x1b[37mfmt\x1b[0m\x1b[37m.\x1b[0m\x1b[36mPrintln\x1b[0m\x1b[37m(\x1b[0m\x1b[33m\"Hello, world!\"\x1b[0m\x1b[37m)\x1b[0m",
```

After adding `light` test case before `dark`, this is the result:

```go
wantOut: "\x1b[0m\x1b[38;5;235mfmt\x1b[0m\x1b[38;5;210m.\x1b[0m\x1b[38;5;35mPrintln\x1b[0m\x1b[38;5;210m(\x1b[0m\x1b[38;5;95m\"Hello, world!\"\x1b[0m\x1b[38;5;210m)\x1b[0m",
```

```go
wantOut: "\x1b[0m\x1b[30mfmt\x1b[0m\x1b[33m.\x1b[0m\x1b[36mPrintln\x1b[0m\x1b[33m(\x1b[0m\x1b[90m\"Hello, world!\"\x1b[0m\x1b[33m)\x1b[0m",
```

Going to step through the debugger a bit to see what's going on.

andyfeller:
@jtmcg @williammartin : Thank you both for feedback while iterating on this PR! 🙇  I'm reopening it after iterating on feedback to make some improvements but also wanting to make the case for certain concerns.

> [!IMPORTANT]
> This PR is dependent on changes in https://github.com/cli/glamour/pull/3 and has been updated to use the latest commit SHA from [`andyfeller/multi-options`](https://github.com/cli/glamour/tree/andyfeller/multi-options).

### Changes

1. Use of new `WithOptions(...TermRendererOption) TermRendererOption` to clean up `WithTheme()`

   The `glamour` TermRendererOption pattern has a limitation that users cannot compose multiple options without duplicating code or building one-off anonymous functions.

   However, this commit takes advantage of an enhancement in https://github.com/cli/glamour/pull/3 that allows users to leverage a helper to avoid building one-offs.  This results in a generic solution to the problem we face with `WithTheme` needing to set multiple options from a single place.

1. Use of new `leaanthony/go-ansi-parser` dependency for parsing ANSI escape sequences and display attributes

   c5336a563c1e02bf13679904649c56740e8718e9 was my initial effort in refactoring the markdown codeblock tests to generically check ANSI escape codes using regular expressions to identify and parse escape sequences.

   While discussing the test with @williammartin, he suggested [`leaanthony/go-ansi-parse`](https://github.com/leaanthony/go-ansi-parser) as an alternative versus building and maintaining our own logic.

1. Expanding `markdown_test.go` to cover accessible color testing while still testing `glamour` and `chroma` output

   There are 2 concerns I have around `glamour` and `chroma` as we implement accessible markdown rendering:

   1. style-appropriate colors are used
   1. 4-bit, base 16 ANSI colors are used when we ask for them.

   In the original PR, only tests concerned with the former existed.  While useful, I agree that they were very situational and maybe a little difficult to grok / maintain.  However, they also highlighted important aspects of `glamour` and `chroma` around registry caching and how if we don't look at these colors we are likely to have low visibility colors used when not appropriate.

   Now, we have a 2nd suite of tests that cover the latter that more generically test ANSI escape sequences for accessibility concerns.

andyfeller:
## Thoughts about testing codeblocks

One of the concerns originally raised around the codeblock color testing was that it was hard to grok and maintain:

https://github.com/cli/go-gh/blob/2b16a4407ce2defc144120dca84f1c40dbb6a8fc/pkg/markdown/markdown_test.go#L153-L178

The concern being the difficulty of reading ANSI escape sequences like:

```go
"\x1b[0m\x1b[38;5;235mfmt\x1b[0m\x1b[38;5;210m.\x1b[0m\x1b[38;5;35mPrintln\x1b[0m\x1b[38;5;210m(\x1b[0m\x1b[38;5;95m\"Hello, world!\"\x1b[0m\x1b[38;5;210m)\x1b[0m"
```

I 💯 agree this isn't ideal and thus experimented with a couple different forms which I'm unsure are legitimately better:

- `fmt.Sprintf(...)`

  From https://github.com/cli/go-gh/pull/183#discussion_r1963838608:

  > ```go
  > 	reset                                            = "\x1b[0m"
  > 	glamourLightCodeBlock_8bitRaisinBlackColorSeq    = "\x1b[38;5;235m"
  > 	glamourLightCodeBlock_8bitVividTangerineColorSeq = "\x1b[38;5;210m"
  > 	glamourLightCodeBlock_8bitJadeColorSeq           = "\x1b[38;5;35m"
  > 	glamourLightCodeBlock_8bitCopperRoseColorSeq     = "\x1b[38;5;95m"
  > ```
  > 
  > ```go
  > fmt.Sprintf("%[1]s%[2]sfmt%[1]s%[3]s.%[1]s%[4]sPrintln%[1]s%[3]s(%[1]s%[5]s\"Hello, world!\"%[1]s%[3]s)%[1]s", reset, glamourLightCodeBlock_8bitRaisinBlackColorSeq, glamourLightCodeBlock_8bitVividTangerineColorSeq, glamourLightCodeBlock_8bitJadeColorSeq, glamourLightCodeBlock_8bitCopperRoseColorSeq)
  > ```

- `text/template`

  ```go
  t, err := template.New("codeblock").Parse(`{{.Reset}}{{.Package}}fmt{{.Reset}}{{.Operator}}.{{.Reset}}{{.Func}}Println{{.Reset}}{{.Operator}}({{.Reset}}{{.Literal}}\"Hello, world!\"{{.Reset}}{{.Operator}}){{.Reset}}`)
  ```
  
  ```go
  data := map[string]string{
      "Reset": "\x1b[0m",
      "Package": "\x1b[38;5;235m",
      "Operator": "\x1b[38;5;210m",
      "Func": "\x1b[38;5;35m",
      "Literal: "\x1b[38;5;95m",
  }
  
  var buf bytes.Buffer
  err = t.Execute(&buf, data)
  if err != nil {
	  panic(err)
  }
  
  result := buf.String()
  ```

I have 2 concerns with the above alternatives:

1. There is significantly more code to understand and reason about
2. Some of these alternatives would affect the design of `Test_Render`, potentially adding to brittleness between test use cases

For these reasons, I kept the original v1 forms of the codeblock tests checking for specific color usage for confidence that reasonably contrasting colors relative to the theme are being displayed.  If there is some other form these can take that is appropriate for testing without being a significant lift, then I appreciate `suggestion`s.

pkg/markdown/markdown.go
```diff
@@ -39,7 +39,10 @@ func WithTheme(theme string) glamour.TermRendererOption {
 		switch theme {
 		case "light", "dark":
 			if accessible {
-				return glamour.WithStyles(AccessibleStyleConfig(theme))
+				return glamour.WithOptions(
```
jtmcg, line 42: This is WAY better ❤️ I think it's super clear what's happening here and I love it. Thanks!

pkg/markdown/markdown_test.go
```diff
@@ -74,14 +150,43 @@ func Test_Render(t *testing.T) {
...
 		t.Run(tt.name, func(t *testing.T) {
+			// Unregister cached chroma style causing codeblock tests to fail based on previous theme
+			delete(styles.Registry, "charm")
 			t.Setenv(accessibility.ACCESSIBILITY_ENV, tt.accessibleEnvVar)
 
-			if tt.styleEnvVar != "" {
-				tmpDir := t.TempDir()
-				path := filepath.Join(tmpDir, fmt.Sprintf("%s.json", tt.styleEnvVar))
+			if tt.styleEnvVar == "" {
+				t.Setenv("GLAMOUR_STYLE", "")
```
jtmcg: Curious as to why this is necessary.

pkg/markdown/markdown_test.go
```diff
@@ -21,12 +24,85 @@ const (
...
+			out, err := Render(tt.text, WithTheme(tt.theme))
+			require.NoError(t, err)
+
+			styledText, err := ansi.Parse(out)
+			require.NoError(t, err)
+
+			for _, st := range styledText {
+				if tt.accessible {
+					require.Equalf(t, st.ColourMode, ansi.Default, "Inaccessible color found in '%s' at %d", st, st.Offset)
+					require.Falsef(t, st.Faint(), "Inaccessible style found in '%s' at %d", st, st.Offset)
```
jtmcg: Can you explain what exactly these two checks are doing? The context for what this is asserting is hidden behind the dependency here. I'd prefer that context to not requiring spelunking into the `ansi` codebase, so maybe a comment explaining how this works would help? 

_Aside:_
I think I have an emerging opinion that a table-test's struct should declare all inputs and outputs. As a corollary to that, I think that means I'm also think it's a bit of a test smell when there's no explicit thing you are expecting in the table test definition. This is not a strongly held opinion right now, but my intuition tells me that makes for easy to understand and easy to extend tests.

pkg/markdown/markdown_test.go
```diff
@@ -74,14 +150,43 @@ func Test_Render(t *testing.T) {
...
+			text:             codeBlock,
+			theme:            "dark",
+			accessibleEnvVar: "true",
+			wantOut:          "\x1b[0m\x1b[37mfmt\x1b[0m\x1b[37m.\x1b[0m\x1b[36mPrintln\x1b[0m\x1b[37m(\x1b[0m\x1b[33m\"Hello, world!\"\x1b[0m\x1b[37m)\x1b[0m",
+		},
 	}
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
+			// Unregister cached chroma style causing codeblock tests to fail based on previous theme
+			delete(styles.Registry, "charm")
```
jtmcg: Consider wrapping this in [`t.Cleanup`](https://pkg.go.dev/testing#T.Cleanup). That should eliminate any artifacts hanging around from these tests after they run instead of cleaning them up before each test case. As it stands right now, I think the last test will leave the cached registry sitting around, which could cause a headache of flakiness down the line.

jtmcg (commented):
A couple of questions and two suggestions, but otherwise this is looking good 👍 I have no further ideas about making the escape codes easier to grok other than what we already discussed, so 🤷

pkg/markdown/markdown.go
andyfeller: Yeah, was a solid call out and will be something others can benefit from.  I appreciate you bringing it up 💪

pkg/markdown/markdown_test.go
andyfeller: The short answer is it was a result of trying to troubleshoot why `glamour` _(or really `chroma` as it turned out)_ was generating incorrect color codes in the testing table.

In trying to reason what was going on within the test relative to this environment variable, I felt having it closer to logic managing `GLAMOUR_STYLE` in the test was easier to reason about but also ensured its value.

I can undo the troubleshooting change

pkg/markdown/markdown_test.go
andyfeller: Heh yeah sorry I didn't want to overthink this test before the PR in case there were cleverer ideas on how to address this test's concerns. 😅 

### What are these checks doing?

These assertions are basically ensuring in an accessible test scenario that:

- escape sequences use 4-bit color
- escape sequences are not using dim/faint formatting

This is a bit of a gray area that I'm taking a more conservative approach than might be necessary:

- some terminals re-render 8-bit and maybe 24-bit colors that correspond to base 16 colors, however that would be terminal dependent so we're saying only 4-bit color is accessible

- initial testing of dim/faint formatting shows decreased color contrast versus the normal and bright versions; its inclusion is an early warning as we build out other examples to raise awareness of its use for further follow up

All of that said, I agree this can be improved with in-line comments and refactoring.

As for your aside 🤔, I generally agree we should have the information in the testing table.  Let me see what I can do rather as my mind is going to hypotheticals that I rather ground in reality.  I got a couple ideas, brb

pkg/markdown/markdown_test.go
andyfeller: Thank you for the suggestion! 🧠

pkg/markdown/markdown_test.go
```diff
@@ -21,12 +24,149 @@ const (
...
+						pageInfo {
+							hasNextPage
+							endCursor
+						}
+					}
+				}
+			}
+		'
+		%[1]s%[1]s%[1]s
+	`)
```
jtmcg, line 62: _Note:_ These big, formatted text blocks give me anxiety in testing 😅 However, as long as we aren't asserting text matching on either of these (which we aren't currently) I feel alright about it. Trying to sort through the ANSI output of this to debug tests would be a nightmare and is why the matching strings I added in `Test_RenderColor` are so short

pkg/markdown/markdown_test.go
jtmcg: Resolved because we ended up with our two testing suites that address these better

pkg/markdown/markdown_test.go
```diff
@@ -74,14 +214,43 @@ func Test_Render(t *testing.T) {
...
+			name:             "when the accessible env var is set and the dark theme is selected, the codeblock renders using 4-bit colors",
+			text:             codeBlock,
+			theme:            "dark",
+			accessibleEnvVar: "true",
+			wantOut:          "\x1b[0m\x1b[37mfmt\x1b[0m\x1b[37m.\x1b[0m\x1b[36mPrintln\x1b[0m\x1b[37m(\x1b[0m\x1b[33m\"Hello, world!\"\x1b[0m\x1b[37m)\x1b[0m",
+		},
 	}
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
+			t.Cleanup(func() {
```
jtmcg, line 246: ❤️

jtmcg (approved):
Great stuff! Thanks for all the back and forth. This really came together 👍

pkg/markdown/markdown_test.go
andyfeller: That is a fair line in the sand :nod:  I tend to value DRYing things up a little where reasonable that doesn't add brittleness.  If these tests change, then I would hope this approach is revisited and changed to fit the evolving need. 💯

PR #186, opened 2025-03-10
Introduce experimental accessible colors in markdown rendering

andyfeller:
Relates github/cli#830

These changes relate to a new experimental feature for rendering markdown in GitHub CLI and CLI extensions using accessible colors.

Out of the box, `markdown` package continues to render GFM using existing `charmbracelet/glamour` light and dark styles in 8-bit color.

By setting `accessible_colors` configuration setting or `GH_ACCESSIBLE_COLORS` environment variable, `markdown` package will begin using ANSI 4-bit colors that most modern terminals allow users to customize.  This will allow users with low vision and color blindness to tailor their experiences more effectively.

## Go version change

This pull request bumps the minimum Go version from 1.21 to 1.23 due to transitive dependencies from experimental Go module:

- https://github.com/charmbracelet/glamour/pull/396
- https://github.com/charmbracelet/glamour/pull/397

## Demo

The following accessible colors have been aligned for consistency with [Primer Design System](https://primer.style/) within the limited palette of 4-bit colors.

### Headers

**Current default**

headers, current release, dark appearance

headers, current release, light apperance

**Accessible**

headers, accessible-colors, dark appearance

headers, accessible-colors, light apperance

**New default**

headers, new default, dark appearance

headers, new default, light appearance

### Horizontal Rule

**Current default**

horizontal rule, current release, dark appearance

horizontal rule, current release, light appearance

**Accessible**

horizontal rule, accessible-colors, dark appearance

horizontal rule, accessible-colors, light appearance

**New default**

horizontal rule, new default, dark appearance

horizontal rule, new default, light appearance

### Images

**Current default**

image, current release, dark appearance

image, current release, light appearance

**Accessible**

image, accessible-colors, dark appearance

image, accessible-colors, light appearance

**New default**

image, new default, dark appearance

image, new default, light appearance

### Links

**Current default**

link, current release, dark appearance

link, current release, light appearance

**Accessible**

link, accessible-colors, dark appearance

link, accessible-colors, light appearance

**New default**

link, new default, dark appearance

link, new default, light appearance

### Code

**Current default**

code, current release, dark appearance

code, current release, light appearance

**Accessible**

code, accessible-colors, dark appearance

code, accessible-colors, light appearance

**New default**

code, new default, dark appearance

code, new default, light appearance

### Quotes

**Current default**

quotes, current release, dark appearance

quotes, current release, light appearance

**Accessible**

quotes, accessible-colors, dark appearance

quotes, accessible-colors, light appearance

**New default**

quotes, new default, dark appearance

quotes, new default, light appearance

### Tables

**Current default**

table, current release, dark appearance

table, current release, light appearance

**Accessible**

table, accessible-colors, dark appearance

table, accessible-colors, light appearance

**New default**

table, new default, dark appearance

table, new default, light appearance


andyfeller:
In addressing the lint errors, I've made minor changes to GitHub Actions workflows to pull Go version from `go.mod`, which requires a newer version of `actions/setup-go`.

Separately while troubleshooting lint job, I notice that the action used for Go linting is 3 major version behind; [`v3.4.0`](https://github.com/golangci/golangci-lint-action/releases/tag/v3.4.0) => [`v6.5.0`](https://github.com/golangci/golangci-lint-action/releases/tag/v6.5.0).

For changes, see [diff](https://github.com/golangci/golangci-lint-action/compare/v3.4.0...v6.5.0)

.github/workflows/lint.yml
```diff
@@ -7,20 +7,21 @@ jobs:
...
           go mod tidy
           git diff --exit-code go.mod
 
       - name: Lint
-        uses: golangci/golangci-lint-action@v3.4.0
+        uses: golangci/golangci-lint-action@v6
         with:
-          version: v1.54
+          version: v1.64
+          problem-matchers: true
```
williammartin, line 27: **Question**

I read the [doc](https://github.com/golangci/golangci-lint-action?tab=readme-ov-file#problem-matchers) but I don't understand what this is doing. Help?

.github/workflows/lint.yml
```diff
@@ -7,20 +7,21 @@ jobs:
...
     steps:
-      - name: Set up Go
-        uses: actions/setup-go@v3
-        with:
-          go-version: "1.21"
-
       - name: Checkout repository
         uses: actions/checkout@v4
 
+      - name: Set up Go
```
williammartin, line 13: **Question**

Why did you reorder the checkout and setup? Same for `lint.yml`

internal/yamlmap/yaml_map.go
```diff
@@ -72,7 +72,7 @@ func (m *Map) AddEntry(key string, value *Map) {
 }
 
 func (m *Map) Empty() bool {
-	return m.Content == nil || len(m.Content) == 0
+	return len(m.Content) == 0
```
williammartin, line 75: **Question**

Can you explain this change please? I'm guessing it was linted because the `nil` slice has `len == 0` but just want to check!

internal/testutils/config_stub.go
```diff
@@ -0,0 +1,21 @@
...
+import (
+	"testing"
+
+	"github.com/cli/go-gh/v2/pkg/config"
+)
+
+// StubConfig replaces the config.Read function with a function that returns a config object
+// created from the given config string. It also sets up a cleanup function that restores the
+// original config.Read function.
+func StubConfig(t *testing.T, cfgStr string) {
```
williammartin, line 12: **Food For Thought**

No concerns merging this as is because it's easy to change later, but maybe this should be `config.Stub` where this lives in `package config_test` would be better than a `testutils` package.

pkg/markdown/accessibility.go
```diff
@@ -0,0 +1,106 @@
...
+	BrightMagenta
+	BrightCyan
+	BrightWhite
+)
+
+func (a ANSIColorCode) String() string {
+	return strconv.Itoa(int(a))
+}
+
+func (a ANSIColorCode) StrPtr() *string {
```
williammartin: **Question**

Did you take this pattern from somewhere?

pkg/markdown/accessibility.go
```diff
@@ -0,0 +1,106 @@
+package markdown
```
williammartin: **Question**

Should this also live under `x` for the moment?

pkg/x/color/accessibility.go
```diff
@@ -0,0 +1,40 @@
...
+	// We are not handling errors because we don't want to fail if the config is not
+	// read. Instead, we assume an empty configuration is equivalent to "disabled".
+	cfg, _ := config.Read(nil)
+	accessibleConfigValue, _ := cfg.Get([]string{AccessibleColorsSetting})
+
+	return accessibleConfigValue == "enabled"
+}
+
+func isEnvVarEnabled(envVar string) bool {
+	return envVar != "" && envVar != "0" && envVar != "false"
```
williammartin: **Food For Thought**

No specific request but this function looks a little sus in that the only place it is used is in this file and repeats the string check that wraps it on line 26.

I think I'd probably inline on 26 like if you wanted to keep the clarity:

```go
if envVar != "" {
   isEnVarEnabled := envVar != "0" && envVar != "false"
   return isEnvVarEnabled
}
```

andyfeller:
> [!NOTE]
> In upping the Go version, there is a discrepancy between automation in the default branch which takes 1.21 into account whereas this branch bumps it up. 🤷 
>
> Meaning there will be some follow up work to the repository ruleset / branch protection rules afterwards.

williammartin (commented):
@andyfeller I got a bit distracted with some other stuff so I didn't get into the good stuff yet in this PR review sorry. Mostly just cursory stuff!

.github/workflows/lint.yml
andyfeller: ### Background

One of the most esoteric features of GitHub Actions is the ability to use annotations, which CI can draw attention to problems or other notable aspects during workflow run.  Using one of the following [Workflow commands](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/workflow-commands-for-github-actions), you are describing where GitHub Actions should display the annotation in the GitHub UI, which pops up as a little box to draw the viewers attention:

- [Setting a notice message](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/workflow-commands-for-github-actions#setting-a-notice-message)

  ```
  ::notice file={name},line={line},endLine={endLine},title={title}::{message}
  ```

- [Setting a warning message](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/workflow-commands-for-github-actions#setting-a-warning-message)

  ```
  ::notice file={name},line={line},endLine={endLine},title={title}::{message}
  ```
  
- [Setting an error message](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/workflow-commands-for-github-actions#setting-an-error-message)

  ```
  ::error file={name},line={line},endLine={endLine},title={title}::{message}
  ```

However, this is a very specific format which people might not like or you might use tools that you don't control but want to bubble up certain problems.  This is where a [**problem matcher**](https://github.com/actions/runner/blob/0ce29d09c64461774c26cd23bc4610d3e94b1edd/docs/adrs/0276-problem-matchers.md#L4) is necessary to identify and translate output from GitHub Actions workflow run into annotations.  I've only found docs on this in the `actions/runner` ADRs; nothing in GitHub Docs.

### `actions/setup-go` problem matcher and `problem-matchers: true`

Now back to this change, `actions/setup-go` sets up a custom problem matcher, however I'm unsure the full story behind the format due to the lack of docs and commentary in https://github.com/actions/setup-go/pull/2:

- [`src/main.ts`](https://github.com/actions/setup-go/blob/5a083d0e9a84784eb32078397cf5459adecb4c40/src/main.ts#L81-L83) sets up a custom problem matcher

  ```typescript
      // add problem matchers
      const matchersPath = path.join(__dirname, '../..', 'matchers.json');
      core.info(`##[add-matcher]${matchersPath}`);
  ```


- [`matchers.json`](https://github.com/actions/setup-go/blob/5a083d0e9a84784eb32078397cf5459adecb4c40/matchers.json) seems like it matches a generic format with nothing like the log-like priority of workflow commands

  ```json
  {
      "problemMatcher": [
          {
              "owner": "go",
              "pattern": [
                  {
                      "regexp": "^\\s*(.+\\.go):(?:(\\d+):(\\d+):)? (.*)",
                      "file": 1,
                      "line": 2,
                      "column": 3,
                      "message": 4
                  }
              ]
          }
      ]
  }
  ```

As for `golangci/golangci-lint-action`, it appears to be designed to conditionally setup problem matchers if `problem-matchers: true` and it has a slightly different setup than `actions/setup-go`:

- [`action.yml`](https://github.com/golangci/golangci-lint-action/blob/b871b4f9de8b829c8d43b0e3bcdf9a8a3f141a3b/action.yml#L45-L48) 

  ```yaml
    problem-matchers:
      description: "Force the usage of the embedded problem matchers"
      default: 'false'
      required: false
  ```

- [`src/run.ts`](https://github.com/golangci/golangci-lint-action/blob/b871b4f9de8b829c8d43b0e3bcdf9a8a3f141a3b/src/run.ts#L68-L77)

  ```typescript
    const problemMatchers = core.getBooleanInput(`problem-matchers`)
  
    if (problemMatchers) {
      const matchersPath = path.join(__dirname, "../..", "problem-matchers.json")
      if (fs.existsSync(matchersPath)) {
        // Adds problem matchers.
        // https://github.com/actions/setup-go/blob/cdcb36043654635271a94b9a6d1392de5bb323a7/src/main.ts#L81-L83
        core.info(`##[add-matcher]${matchersPath}`)
      }
    }
  ```

- [`problem-matchers.json`](https://github.com/golangci/golangci-lint-action/blob/b871b4f9de8b829c8d43b0e3bcdf9a8a3f141a3b/problem-matchers.json)

  ```json
  {
    "problemMatcher": [
      {
        "owner": "golangci-lint-colored-line-number",
        "severity": "error",
        "pattern": [
          {
            "regexp": "^([^:]+):(\\d+):(?:(\\d+):)?\\s+(.+ \\(.+\\))$",
            "file": 1,
            "line": 2,
            "column": 3,
            "message": 4
          }
        ]
      }
    ]
  }
  ```

### Which should we use?

1. `actions/setup-go`

   ```json
   "regexp": "^\\s*(.+\\.go):(?:(\\d+):(\\d+):)? (.*)"
   ```

1. `golangci/golangci-lint-action` with an `error` severity

   ```json
   "regexp": "^([^:]+):(\\d+):(?:(\\d+):)?\\s+(.+ \\(.+\\))$"
   ```

Between the 2, I would assume the problem matcher setup by `golangci` is more appropriate for its use than the generic, non-severity problem matcher from `actions/setup-go`.

the end

.github/workflows/lint.yml
andyfeller: In order for `actions/setup-go@v5` to use the `go.mod` for determining the Go toolchain version.  These workflows having a hardcoded Go version makes situations where the version changes a little more painful than needed.

internal/yamlmap/yaml_map.go
andyfeller: Great question!  This was the only error / warning raised by `golangci-lint`:

```shell
$ golangci-lint run --out-format=colored-line-number
internal/yamlmap/yaml_map.go:75:9: S1009: should omit nil check; len() for nil slices is defined as zero (gosimple)
	return m.Content == nil || len(m.Content) == 0
	       ^
```

pkg/markdown/accessibility.go
andyfeller: Yeah, I believe it was based on 2 ideas:

- [`charmbracelet/glamour` default styles logic](https://github.com/charmbracelet/glamour/blob/4c040b7fd5c023154de93d5c0f789111ea06c82c/styles/styles.go#L135-L343):

  ```go
	  // DarkStyleConfig is the default dark style.
	  DarkStyleConfig = ansi.StyleConfig{
		  Document: ansi.StyleBlock{
			  StylePrimitive: ansi.StylePrimitive{
				  BlockPrefix: "\n",
				  BlockSuffix: "\n",
				  Color:       stringPtr("252"),
			  },
			  Margin: uintPtr(defaultMargin),
		  },
		  BlockQuote: ansi.StyleBlock{
			  StylePrimitive: ansi.StylePrimitive{},
			  Indent:         uintPtr(1),
			  IndentToken:    stringPtr("│ "),
		  },
  ```
  
  which uses a handful of [helper functions](https://github.com/charmbracelet/glamour/blob/4c040b7fd5c023154de93d5c0f789111ea06c82c/styles/styles.go#L678-L680):
  
  ```go
  func boolPtr(b bool) *bool       { return &b }
  func stringPtr(s string) *string { return &s }
  func uintPtr(u uint) *uint       { return &u }
  ```

- [`color.go` in `cli/cli` and `ansi.ColorFunc()`](https://github.com/cli/cli/blob/af4acb380136fd106b38cc0ab404ff975bca9795/pkg/iostreams/color.go#L15-L33)

  This was more inspirational about having a type where information about color could then be leveraged elsewhere rather than the unexported private functions above:

  ```go
  var (
	  magenta            = ansi.ColorFunc("magenta")
	  cyan               = ansi.ColorFunc("cyan")
	  red                = ansi.ColorFunc("red")
	  yellow             = ansi.ColorFunc("yellow")
	  blue               = ansi.ColorFunc("blue")
	  green              = ansi.ColorFunc("green")
	  gray               = ansi.ColorFunc("black+h")
	  lightGrayUnderline = ansi.ColorFunc("white+du")
	  bold               = ansi.ColorFunc("default+b")
	  cyanBold           = ansi.ColorFunc("cyan+b")
	  greenBold          = ansi.ColorFunc("green+b")
	  highlightStart     = ansi.ColorCode(highlightStyle)
	  highlight          = ansi.ColorFunc(highlightStyle)
  
	  gray256 = func(t string) string {
		  return fmt.Sprintf("\x1b[%d;5;%dm%s\x1b[m", 38, 242, t)
	  }
  )
  ```
```

File: pkg/markdown/accessibility.go (hunk 6 above)
Comment (andyfeller) [reply pkg/markdown/accessibility.go]: Great question!  I can get behind that especially as we don't want anyone thinking its stable, yeah.

File: pkg/x/color/accessibility.go (hunk 7 above)
Comment (andyfeller) [reply pkg/x/color/accessibility.go]: Totally fair call out!

I was hoping we could tie into existing logic as this work evolved as this shouldn't be a new problem, but I suppose we don't really:

https://github.com/cli/go-gh/blob/7a6e5543a7f13a372dfafa88fac99981496b4991/pkg/api/http_client.go#L80-L88

Comment (andyfeller) [reply pkg/x/color/accessibility.go]: We probably also want to add `no`, too

File: .github/workflows/lint.yml (hunk 1 above)
Comment (williammartin) [reply .github/workflows/lint.yml:27]: wat

Comment (williammartin) [reply .github/workflows/lint.yml:27]: Thanks for the explanation. I guess at some point it would be good to see a comparison of the two visually.

File: .github/workflows/lint.yml (hunk 2 above)
Comment (williammartin) [reply .github/workflows/lint.yml:13]: Got it, thanks!

File: internal/yamlmap/yaml_map.go (hunk 3 above)
Comment (williammartin) [reply internal/yamlmap/yaml_map.go:75]: Cool, thanks.

File: pkg/markdown/accessibility.go (hunk 5 above)
Comment (williammartin) [reply pkg/markdown/accessibility.go]: Alright, makes sense to me.

Maybe it doesn't need to be exported?

File: pkg/x/color/accessibility_test.go [hunk 8]
```diff
@@ -0,0 +1,92 @@
...
+		{
+			name:        "When the accessibility configuration is set to enabled and the env var is set to 'false', it should return false",
+			envVarValue: "false",
+			cfgStr:      accessibilityEnabledConfig(),
+			wantOut:     false,
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			t.Setenv("GH_ACCESSIBLE_COLORS", tt.envVarValue)
```
williammartin: **Food For Thought**

In practice I think this will work but I'm not sure it's totally correct in line with the descriptions e.g. the first test says:

```
When the accessibility configuration and env var are both unset, it should return false
```

But setting an env var to an empty string is not the same as it being unset. You can confirm this with:

```go
			_, set := os.LookupEnv("GH_ACCESSIBLE_COLORS")
			fmt.Println(set)
```

williammartin: Actually if you look at this test:

```go
		{
			name:        "When the accessibility configuration is set to disabled and the env var is unset, it should return false",
			envVarValue: "",
			cfgStr:      accessibilityDisabledConfig(),
			wantOut:     false,
		},
```

You might expect that changing the implementation of `IsAccessibleColorsEnabled` with a refactor to a true "set" or "unset" check would be safe:

```go
	envVar, set := os.LookupEnv(AccessibleColorsEnv)
	if set {
		return isEnvVarEnabled(envVar)
	}
```

But it's not because `""` is considered set. This refactor actually results in the config **never** being checked in the tests.

For me, this is enough to **Request Changes** to make it clear what an empty string vs unset should do.

pkg/x/color/accessibility_test.go
```diff
@@ -0,0 +1,92 @@
...
+			testutils.StubConfig(t, tt.cfgStr)
+			assert.Equal(t, tt.wantOut, IsAccessibleColorsEnabled())
+		})
+	}
+}
+
+func accessibilityEnabledConfig() string {
+	return heredoc.Docf(`
+		%s: enabled
+	`, AccessibleColorsSetting)
```
williammartin: **Nit**

I understand your desire to DRY out the use of this config but I have mixed feelings about it. The thing is, the tests are supposed to act as guardrails but changing the value of `AccessibleColorsSetting` would break users but wouldn't fail the tests.

I think it's generally better to duplicate in tests for this reason.

williammartin: In fact you avoided this trap for the envvar by using the string directly in the test.

pkg/markdown/accessibility_test.go
```diff
@@ -0,0 +1,118 @@
...
+
+import (
+	"testing"
+
+	"github.com/charmbracelet/glamour/ansi"
+	"github.com/charmbracelet/glamour/styles"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestANSIColorCodeString(t *testing.T) {
```
williammartin: **Question**

Maybe this is pedantic but it feels like we should test all the colors here. I swapped the order of some of them and no tests failed.

Maybe it's unnecessary, who's really going to swap the order? Maybe a comment would be fine. I don't think it's obvious to uninformed readers that the order of the enumeration is important.

An alternative would be to use a map to `int`.

What do you think?

pkg/markdown/markdown.go
```diff
@@ -33,9 +34,16 @@ func WithWrap(w int) glamour.TermRendererOption {
 // If the environment variable GLAMOUR_STYLE is set, it will take precedence over the provided theme.
 func WithTheme(theme string) glamour.TermRendererOption {
 	style := os.Getenv("GLAMOUR_STYLE")
+	accessible := color.IsAccessibleColorsEnabled()
```
williammartin: **Nit**

If this is only used inside the `light`, `dark` case then maybe just inline it in the `if` there?

It would probably be a totally negligible performance improvement but generally I value the scoping being tight.

pkg/markdown/markdown_test.go
```diff
@@ -1,38 +1,282 @@
...
+	glamourDarkH2_8bitColorSeq  = "\x1b[38;5;39;"
+	customH2_8bitColorSeq       = "\x1b[38;5;61;"
+	magenta_4bitColorSeq        = "\x1b[35;"
+	brightMagenta_4bitColorSeq  = "\x1b[95;"
+)
+
+// Test_RenderAccessible tests rendered markdown for accessibility concerns such as color mode / depth and other display attributes.
+// It works by parsing the rendered markdown for ANSI escape sequences and checking their display attributes.
+// Test scenarios allow multiple color mode / depths because `ansi.Parse()` considers `\x1b[0m` sequence as part of `ansi.Default`.
+func Test_RenderAccessible(t *testing.T) {
```
williammartin, line 30: **Praise**

This test is sick.

pkg/markdown/markdown_test.go
```diff
@@ -1,38 +1,282 @@
...
+			accessibleEnvVar: "true",
+			wantOut:          "\x1b[0m\x1b[37mfmt\x1b[0m\x1b[37m.\x1b[0m\x1b[36mPrintln\x1b[0m\x1b[37m(\x1b[0m\x1b[33m\"Hello, world!\"\x1b[0m\x1b[37m)\x1b[0m",
 		},
 	}
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
-			_, err := Render(tt.text, WithTheme(tt.theme))
-			assert.NoError(t, err)
+			t.Cleanup(func() {
+				// Chroma caches charm style used to render codeblocks, it must be unregistered to avoid previously used style being reused.
```
williammartin, line 319: **Praise**

Great comment.

pkg/markdown/markdown_test.go
```diff
@@ -1,38 +1,282 @@
...
+				delete(styles.Registry, "charm")
+			})
+			t.Setenv(color.AccessibleColorsEnv, tt.accessibleEnvVar)
```
williammartin: **Nit**

Same issue as DRYing the config setting mentioned in another comment.

pkg/markdown/markdown_test.go
```diff
@@ -1,38 +1,282 @@
...
+
+			if tt.styleEnvVar != "" {
+				path := filepath.Join(t.TempDir(), fmt.Sprintf("%s.json", tt.styleEnvVar))
+				err := os.WriteFile(path, []byte(customGlamourStyle(t)), 0644)
+				if err != nil {
+					t.Fatal(err)
```
williammartin: **Question**

Is there a reason we're not using `require.NoError(err)` here? Suspect not.

williammartin (changes requested):
Lots of good stuff in here, don't think there's too much work to resolve my comments.

I haven't yet pulled this into `cli/cli` to use it.

.github/workflows/lint.yml
andyfeller: Additional documentation on problem matchers:
https://github.com/actions/toolkit/blob/7ae578ddd11b9e33db1daaad2ffb39e1ffeada3f/docs/problem-matchers.md

pkg/x/color/accessibility_test.go
andyfeller: Makes sense!

pkg/x/color/accessibility_test.go
andyfeller: Ooh good catch!

pkg/markdown/accessibility_test.go
andyfeller: There isn't a comment in https://github.com/cli/go-gh/pull/182 but I recall at the time bringing up some past experiences relying upon enum and iota in other codebases.  Now this has been brought up completely out of the blue, I'm more confident maybe this changes.

pkg/markdown/markdown_test.go
andyfeller: I'm unsure and comfortable changing it.

What are situations you would expect `t.Fatal(...)` being used?

williammartin: None, if you're using `testify` across the codebase. If you're not using an assertion library, then I would expect to see it.

internal/testutils/config_stub.go
andyfeller: I'm down for this, but would probably like to sync over a weird behavior around importing the package I'm seeing while acting on this. 🤔

pkg/markdown/accessibility.go
andyfeller: If we're moving this under `x`, don't we have to export it?

internal/testutils/config_stub.go
williammartin: On second thoughts it might not be possible to import from a different package for `_test` tags. Sorry for the wild goose chase.

pkg/markdown/accessibility.go
williammartin: It kind of looked like `StrPtr` was only used in this package. Am I missing something?

pkg/markdown/accessibility.go
andyfeller: Related question: any preference on the alias used when importing `x` packages?

I know there are broader Go conventions on the topic, but I don't think I've encountered another project that ran into this situation to have a strong opinion and thus went with:

```go
// Package markdown facilitates rendering markdown in the terminal.
package markdown

import (
	"os"
	"strings"

	"github.com/charmbracelet/glamour"
	xcolor "github.com/cli/go-gh/v2/pkg/x/color"
	xmarkdown "github.com/cli/go-gh/v2/pkg/x/markdown"
)
```

williammartin: No opinions.

pkg/markdown/accessibility.go
andyfeller: No you aren't 🚀 let's unexport it.

pkg/markdown/accessibility_test.go
andyfeller: So after a bunch of back and forth, here are my thoughts:

1. The enumeration is the only way we can really do compile time type checking, so will leave it for now
2. I think ANSI color code is a little misleading _(technically ANSI colors = 3-bit)_
3. I don't know if we need separate `String` and `StrPtr` functions
4. I think renaming this to explain its purpose might be easier for people to understand; basically these are glamour style color code constants
5. I agree we should test all the colors

I pushed some changes up and happy to talk through them.

pkg/x/markdown/accessibility_test.go
```diff
@@ -0,0 +1,165 @@
...
+			want: "14",
+		},
+		{
+			name: "bright white",
+			c:    brightWhite,
+			want: "15",
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
```
williammartin, line 101: **Suggestion**

Mark the subtests with `t.Parallel()` since there's nothing that prevents them being parallelised. Same for all the other tests.

williammartin (approved):
### `./bin/gh repo view` with and without `GH_ACCESSIBLE_COLORS=1`

image showing markdown rendering

And on a white background with `GH_ACCESSIBLE_COLORS=1`

image

### With config setting

```
➜  cli git:(trunk) ✗ gh config set accessible_colors enabled
! warning: 'accessible_colors' is not a known configuration key
```

image

williammartin:
@andyfeller There is text in the images above "Image: screenshot of gh pr status → https://user-images.githubusercontent.com/98482/84171218-327e7a80-aa40-11ea-8cd1-5177fc2d0e72.png" and I'm having trouble convincing myself that it is using ANSI 4-bit colors.

image

From using `sequin` it looks like the control code is:

```
 CSI 37m: Foreground color: White
 CSI 38;5;243m: Foreground color: 243
```

So this looks like a 256-bit color code. I think this is the relevant style bits are:
 * https://github.com/charmbracelet/glamour/blob/d1d5125cfab42203495f15b846e901d0c1f37aaa/styles/styles.go#L233
 * https://github.com/charmbracelet/glamour/blob/d1d5125cfab42203495f15b846e901d0c1f37aaa/styles/styles.go#L442

---

It also looks to me as if [LinkText](https://github.com/charmbracelet/glamour/blob/d1d5125cfab42203495f15b846e901d0c1f37aaa/styles/styles.go#L434) would not be 16-bit, but I could be mistaken.

Here's the green `LinkText` "hub" on my dark background:

image

Indeed, when I run it through `sequin` I see:

```
 CSI 0m: Reset style
 CSI 38;5;35;1m: Foreground color: 35, Bold
Text more detailed explanation
```

Which matches the `DarkStyleConfig` [LinkTest](https://github.com/charmbracelet/glamour/blob/d1d5125cfab42203495f15b846e901d0c1f37aaa/styles/styles.go#L224-L226)

williammartin (changes requested):
Requesting changes based on ^

andyfeller:
## Accessible colors should also be used when `GLAMOUR_STYLE` is `dark` or `light`

### Problem

When determining how to render markdown, the code is concerned with 2 things:

1. Is the terminal background `dark` or `light`?  Or should `none` colors be used? _(`theme`)_ 

2. How does the user want to stylize markdown? _(`style`)_
   - Using a built-in [`charmbracelet/glamour` style](https://github.com/cli/go-gh/blob/7a6e5543a7f13a372dfafa88fac99981496b4991/pkg/term/env.go#L139-L148) _(`ascii`, `auto`, `dark`, `dracula`, `tokyo-night`, `light`, `notty`, `pink`)_
   - Using a path to a [custom `charmbracelet/glamour` style](https://github.com/charmbracelet/glamour/tree/master?tab=readme-ov-file#styles) defined as JSON

Currently, **accessible colors are only applied if `style` has not been set**, which can be confusing if the user sets `GLAMOUR_STYLE` to `dark` or `light`:

https://github.com/cli/go-gh/blob/bde34d5cca49d621bce438a47b3ab94c4246856a/pkg/markdown/markdown.go#L33-L53

- `GH_ACCESSIBLE_COLORS=1 bin/gh repo view andyfeller/markdown-lorem-ipsum`

  As we see, `gh` is using the custom accessible themes when `GLAMOUR_STYLE` is unset:

  Screenshot of new accessible experience without glamour env variable

- `GLAMOUR_STYLE=dark GH_ACCESSIBLE_COLORS=1 bin/gh repo view andyfeller/markdown-lorem-ipsum`

  However we see `gh` will fallback to the current markdown rendering experience even when the `style` and `theme` should be the same.

  Screenshot of new accessible experience with glamour env variable

### Suggestion

This was honestly a bit confusing for me and I've been working on these changes for a while as I expected `GLAMOUR_STYLE=dark` or `GLAMOUR_STYLE=light` to also pick up accessible colors.

If this logic was enhanced to allow those 2 `styles` to fall into the same logic that applies accessible colors, I think we'll have fewer people surprised.

andyfeller:
## Should tables be wrapped or not? 🤔 

One of the improvements in `charmbracelet/glamour@v0.9.0` is [greater control over table wrapping / overflow](https://github.com/charmbracelet/lipgloss/releases/tag/v1.1.0), which `glamour` now [does for tables by default](https://github.com/charmbracelet/glamour/blob/f43b1ad9ef09b10a93837e07ae2c18b66bceac5c/ansi/table.go#L35-L61):

```go
	ctx.table.lipgloss = table.New().Width(width).Wrap(true)
```

**UPDATE: This PR has been updated with upstream changes to disable table content wrapping until we can understand its impact to screenreader accessibility.**

<details>
<summary>
  <b>Expand for full thoughts</b>
</summary>
<p>

I'm going to enhance https://github.com/andyfeller/markdown-lorem-ipsum to contain an example for discussion as I'm unsure how this affects accessible screenreader experience, which is outside the scope of the issue and PR here.

### Current `gh` markdown table rendering experience

Screenshot 2025-03-13 at 5 54 04 PM

### New `gh` experience table rendering experience

Screenshot 2025-03-13 at 5 53 14 PM
Screenshot 2025-03-13 at 5 53 38 PM


cc: @williammartin 


</p>
</details>

pkg/x/markdown/accessibility_test.go
andyfeller: TIL `t.Parallel()` cannot be used with `t.Setenv()`

That said, I got as many as I could.

andyfeller:
@williammartin : I'm going to spend some time later tonight putting together a demo of the changes in the PR just so we can confirm the experience meets expectations beyond the tests and code review.  Though the fix for table wrapping hasn't been released, I think it's okay we pin to the commit for now to get this merged.

So re-re-requesting review and I'll follow up on demoing it later tonight.

**UPDATE: See PR body above for updated screenshots**

andyfeller:
Want to give a special shout out to our friends at [`charmbracelet`](https://github.com/charmbracelet/) who helped make this work possible with work on `glamour` and `lipgloss` in no specific order:

- @andreynering
- @aymanbagabas
- @bashbunni
- @caarlos0
- @meowgorithm

go.mod
```diff
@@ -1,27 +1,31 @@
 module github.com/cli/go-gh/v2
 
-go 1.21
+go 1.23.0
+
+toolchain go1.23.7
```
williammartin: Is this really required? Would think it is unusual to have a toolchain entry for a library.

williammartin (approved):
Approved with one non blocking comment! Well done on all this, I know it was a lot.

go.mod
andyfeller: Great question; `go mod` did this in upgrading dependencies 🤦

andyfeller: Removing for now

PR #200, opened 2025-10-31
Upgrade Golangci-lint to `v2.6`

### LINKED ISSUE #199: Upgrade Golangci-lint to `v2.6`
This is to keep the Golangci-lint version in sync with `gh` codebase.

babakks:
Fixes #199

Also, fixed godocs as suggested by `godoclint`, but didn't enable the linter.

### DIFF
--- a/pkg/term/env_test.go
+++ b/pkg/term/env_test.go
@@ -1,5 +1,3 @@
-// Package term provides information about the terminal that the current process is connected to (if any),
-// for example measuring the dimensions of the terminal and inspecting whether it's safe to output color.
 package term
 
 import (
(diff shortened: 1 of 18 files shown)


pkg/term/env_test.go
```diff
@@ -1,5 +1,3 @@
-// Package term provides information about the terminal that the current process is connected to (if any),
-// for example measuring the dimensions of the terminal and inspecting whether it's safe to output color.
 package term
```
babakks, line 1: This is redundant since this is a test package, and also the doc is present in `pkg/term/env.go`:

https://github.com/cli/go-gh/blob/f6d1f605e343beb2b8d89fb0440b95a4554db3c1/pkg/term/env.go#L1-L3

BagToad (approved):

PR #209, opened 2026-01-25
config: allow overriding data directory via `GH_DATA_DIR`

amaanq:
Hey, this is a bit of a drive-by PR but I think it's quite small in nature and fills in a gap that'd help with consistency. I've added support for overriding the data directory via a `GH_DATA_DIR` environment variable, which is akin to `GH_CONFIG_DIR`. This is nice for users who want a global/user-agnostic configuration, and don't want to pollute their home dirs. In my case, I want to manage extensions in a different directory than the default (I use NixOS and want to have my config be read-only in `/etc/gh`).

The precedence for locating the data dir is now `GH_DATA_DIR` --> `XDG_DATA_HOME` --> `LocalAppData` (Windows) --> `HOME`

While I was at it, I also fixed tests that suffered from environment pollution. I set `GH_TOKEN` and this was breaking some tests.

Thanks a ton!

N.B. If accepted, I can help update the main CLI repo with docs :)

### DIFF
--- a/pkg/api/client_options_test.go
+++ b/pkg/api/client_options_test.go
@@ -9,6 +9,9 @@ import (
 )
 
 func TestResolveOptions(t *testing.T) {
+	t.Setenv("GH_CONFIG_DIR", "")
+	t.Setenv("GH_TOKEN", "")
+	t.Setenv("GITHUB_TOKEN", "")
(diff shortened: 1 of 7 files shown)


kwanawyly911-svg (commented):
Correct

PR #211, opened 2026-01-27
Adding application/octocat-stream to list of inspectable MIME types

### LINKED ISSUE #210: Treat application/octocat-stream as an inspectable MIME type
Relates to: https://github.com/github/rest-api-description/issues/5178

When using the GitHub CLI with --verbose flag on endpoints that return non-standard content types (like application/octocat-stream from /octocat), the response body is not displayed even though the request succeeds.

```
$ gh api --verbose /octocat
* Request at 2026-01-27 15:15:09.70376 -0600 CST m=+0.070954459
* Request to https://api.github.com/octocat
> GET /octocat HTTP/1.1
> Host: api.github.com
...
< HTTP/2.0 200 OK
< Content-Type: application/octocat-stream
...

# No response body shown! 😿
```

Meanwhile, gh api /octocat (without --verbose) works correctly and displays the octocat ASCII art. The same applies to other tools like `curl` and `wget`.

## Root Cause

The [inspectableMIMEType](https://github.com/cli/go-gh/blob/a0a6e8947ae2ceedb496654757886ef41ef5ac72/pkg/api/http_client.go#L122-L125) determines which response bodies get logged. This function is used in the httpretty logger's [body filter](https://github.com/cli/go-gh/blob/a0a6e8947ae2ceedb496654757886ef41ef5ac72/pkg/api/http_client.go#L105-L107). Since `application/octocat-stream` doesn't match any of these patterns, the body gets filtered out.

## Proposed fix

Add `application/octocat-stream` to the list of inspectable MIME types

shawnHartsell:
Fixes: #210
Relates to: https://github.com/github/rest-api-description/issues/5178  

This pull request updates the logic for determining which MIME types are considered "inspectable" and adds comprehensive unit tests to ensure correct behavior. The main change is the inclusion of the `application/octocat-stream` MIME type as inspectable.

I also manually tested this change by replacing the `cli/go-gh` package for cli/cli locally and verified that running `gh api --verbose /octocat` output the ASCII art response.

### DIFF
--- a/pkg/api/http_client.go
+++ b/pkg/api/http_client.go
@@ -122,6 +122,7 @@ func NewHTTPClient(opts ClientOptions) (*http.Client, error) {
 func inspectableMIMEType(t string) bool {
 	return strings.HasPrefix(t, "text/") ||
 		strings.HasPrefix(t, "application/x-www-form-urlencoded") ||
+		strings.HasPrefix(t, "application/octocat-stream") ||
 		jsonTypeRE.MatchString(t)
 }
 
--- a/pkg/api/http_client_test.go
+++ b/pkg/api/http_client_test.go
@@ -184,3 +184,59 @@ func printPendingMocks(mocks []gock.Mock) string {
 	}
 	return fmt.Sprintf("%d unmatched mocks: %s", len(paths), strings.Join(paths, ", "))
 }
+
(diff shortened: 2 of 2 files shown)


kwanawyly911-svg (commented):
The

PR #213, opened 2026-02-13
Pin REST API version to 2022-11-28

williammartin:
## Description

Currently, `go-gh` does not set the `X-GitHub-Api-Version` header, meaning if/when a new version rolls out, consumers of `go-gh` using APIs that have breaking changes may begin breaking.

This adds `X-GitHub-Api-Version: 2022-11-28` as a default header in `resolveHeaders()`, following the same pattern as the other defaults (Content-Type, User-Agent, etc.). Consumers can override it via `ClientOptions.Headers`, and `SkipDefaultHeaders: true` excludes it as expected.

See: https://github.blog/developer-skills/github/to-infinity-and-beyond-enabling-the-future-of-githubs-rest-api-with-api-versioning/

### DIFF
--- a/pkg/api/http_client.go
+++ b/pkg/api/http_client.go
@@ -22,6 +22,8 @@ import (
 const (
 	accept          = "Accept"
 	authorization   = "Authorization"
+	apiVersion      = "X-GitHub-Api-Version"
+	apiVersionValue = "2022-11-28"
 	contentType     = "Content-Type"
 	github          = "github.com"
 	jsonContentType = "application/json; charset=utf-8"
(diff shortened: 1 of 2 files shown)


pkg/api/http_client.go
BagToad: Completely unrelated to this PR but I'd love this to be named `setDefaultHeaders` instead because `resolveHeaders` I feel is a little misleading for how this is actually used - it's only called to set default headers.

BagToad (approved):
LGTM

pkg/api/http_client.go
williammartin: Done.

PR #214, opened 2026-02-20
Remove dependency on archived github.com/cli/safeexec

matheuscscp:
The safeexec package was a workaround for a security issue in exec.LookPath on Windows, where the current working directory was searched before PATH entries. This was fixed in Go 1.19 (see golang/go#43724), making safeexec redundant.

Since go-gh requires Go 1.25, replace all safeexec.LookPath calls with exec.LookPath from the standard library.

### DIFF
--- a/gh.go
+++ b/gh.go
@@ -12,8 +12,6 @@ import (
 	"io"
 	"os"
 	"os/exec"
-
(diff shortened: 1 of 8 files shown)


matheuscscp:
Bump

PR #216, opened 2026-03-01
Strip rate-limit headers from cached responses

kim-em:
## Summary

After rate limit exhaustion, cached GraphQL responses carry stale `X-Ratelimit-Remaining: 0` headers. Since the cache TTL (up to 24h) far exceeds the rate limit reset window (~1h), callers reading these headers from a cached response incorrectly believe the rate limit is still exhausted.

This causes `gh issue list` and similar commands to fail for hours after the actual rate limit has reset, while `gh api graphql` (which bypasses the cache) works fine.

This PR strips `X-Ratelimit-*` headers when serving responses from cache, since they reflect the state at the original request time and are not meaningful for cached responses.

- Strips `X-Ratelimit-{Limit,Remaining,Reset,Used,Resource}` headers in `fileStorage.read()`
- Adds a test verifying rate-limit headers are stripped from cached responses while other headers are preserved

Fixes cli/cli#12812

🤖 Prepared with Claude Code

### DIFF
--- a/pkg/api/cache.go
+++ b/pkg/api/cache.go
@@ -167,6 +167,19 @@ func (fs *fileStorage) read(key string) (*http.Response, error) {
 	}
 
 	res, err := http.ReadResponse(bufio.NewReader(body), nil)
+	if err == nil {
+		// Strip rate-limit headers from cached responses. These headers
+		// reflect the state at the time the response was originally received
+		// and become stale long before the cache TTL expires (rate limits
+		// reset hourly, but cache TTL can be up to 24h). Serving stale
+		// rate-limit headers causes callers to believe the rate limit is
+		// still exhausted after it has already reset.
+		res.Header.Del("X-Ratelimit-Limit")
+		res.Header.Del("X-Ratelimit-Remaining")
+		res.Header.Del("X-Ratelimit-Reset")
+		res.Header.Del("X-Ratelimit-Used")
+		res.Header.Del("X-Ratelimit-Resource")
+	}
 	return res, err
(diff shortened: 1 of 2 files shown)


Eviphon:
> ## Summary
> After rate limit exhaustion, cached GraphQL responses carry stale `X-Ratelimit-Remaining: 0` headers. Since the cache TTL (up to 24h) far exceeds the rate limit reset window (~1h), callers reading these headers from a cached response incorrectly believe the rate limit is still exhausted.
> 
> This causes `gh issue list` and similar commands to fail for hours after the actual rate limit has reset, while `gh api graphql` (which bypasses the cache) works fine.
> 
> This PR strips `X-Ratelimit-*` headers when serving responses from cache, since they reflect the state at the original request time and are not meaningful for cached responses.
> 
> * Strips `X-Ratelimit-{Limit,Remaining,Reset,Used,Resource}` headers in `fileStorage.read()`
> * Adds a test verifying rate-limit headers are stripped from cached responses while other headers are preserved
> 
> Fixes [cli/cli#12812](https://github.com/cli/cli/issues/12812)
> 
> 🤖 Prepared with Claude Code

Eviphon:
kim-em:fix-stale-ratelimit-cache-headers

niik:
Hey @kim-em. We typically have automation to close PRs that don't reference an issue with a help-wanted label but that doesn't work in this cross-repo situation so I'm gonna close this manually. Let's stick to the issue until we reach a conclusion on how we want to resolve this!

PR #263, opened 2026-07-19
fix(api): publish HTTP cache entries without in-place writes

### LINKED ISSUE #252: API cache is not write-atomic.
The cache's [`store`](https://github.com/cli/go-gh/blob/a89b68a441b259f483266e4cee718d9a8e606722/pkg/api/cache.go#L173) method is not atomic. It is protected by a mutex, but this mutex only prevents concurrent in-process writes. If multiple invocations of the `gh` CLI are run in parallel then two concurrent writes to the same cache file can occur, corrupting the cache.

Similarly if a process dies mid-write then a partially written file can end up in the cache.

A simple fix for this would be to make the cache write atomic by writing to a temporary file and then doing an atomic move to overwrite the cache entry.

baiyuxi930826:
Fixes #252.

### Description

go-gh caches API responses on disk so later requests can reuse them. Previously, saving a response truncated the existing cache file before writing its replacement. An in-process mutex cannot coordinate separate `gh` processes, so another process could read an incomplete response or overwrite the same file concurrently. A failed or interrupted write could also destroy a usable entry.

Write the response to a temporary file in the same directory, close it successfully, and then rename it into place. Failed response writes leave the existing entry untouched. On Windows, briefly retry file-sharing conflicts before giving up on caching; the caller still receives the API response.

Not tested manually. I exercised this library change only through automated tests; I did not build and run `gh` against this branch.

### Key points

- Same-directory rename provides atomic replacement on Unix. Go does not guarantee atomic rename on Windows or other non-Unix platforms; retries improve handling of Windows contention, not that guarantee.
- Windows sharing violations and potentially transient access-denied errors use exponential backoff within a 100 ms retry budget. Other errors return immediately. There is no fallback to truncating or copying over the destination.
- Keep the existing in-process mutex without adding cross-process locks or custom Windows file-opening semantics. Caching remains best-effort.
- Deferred close and removal clean up temporary files on errors and body panics. Forced process termination can still leave an unpublished temporary file behind.
- No file or directory `fsync` is added: this cache does not promise durability across power loss or operating-system crashes.

### Notes for reviewers

Start with `fileStorage.store` in `pkg/api/cache.go`, then the rename helper and its Windows error classification in `pkg/api/cache_rename*.go`. The publication and rename tests describe the failure, concurrency, and retry scenarios.

[Issue #252](https://github.com/cli/go-gh/issues/252) supplies the cache-corruption context motivating the change and we also saw it in https://github.com/cli/cli/issues/14394

### Authorship and follow-up

Who wrote this:

- [ ] A human wrote it.

The original patch is by @baiyuxi930826. Copilot wrote the follow-up implementation and this description under human direction.

Who answers review comments:

- [x] @williammartin will read and reply directly.

### DIFF
--- a/pkg/api/cache.go
+++ b/pkg/api/cache.go
@@ -176,18 +176,21 @@ func (fs *fileStorage) store(key string, res *http.Response) (storeErr error) {
 	fs.mu.Lock()
 	defer fs.mu.Unlock()
 
-	if storeErr = os.MkdirAll(filepath.Dir(cacheFile), 0755); storeErr != nil {
+	dir := filepath.Dir(cacheFile)
+	if storeErr = os.MkdirAll(dir, 0755); storeErr != nil {
 		return
 	}
 
+	// Write to a temporary file in the same directory, then rename into place
+	// so concurrent processes and partial writes cannot corrupt the cache entry.
 	var f *os.File
-	if f, storeErr = os.OpenFile(cacheFile, os.O_WRONLY|os.O_CREATE|os.O_TRUNC, 0600); storeErr != nil {
+	if f, storeErr = os.CreateTemp(dir, ".gh-cache-*"); storeErr != nil {
 		return
 	}
-
+	tmpName := f.Name()
 	defer func() {
-		if err := f.Close(); storeErr == nil && err != nil {
-			storeErr = err
+		if storeErr != nil {
+			_ = os.Remove(tmpName)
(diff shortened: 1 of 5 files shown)


niik (approved):

williammartin:
Thanks @baiyuxi930826, I used your work as a base to make some improvements.

PR #274, opened 2026-07-30
fix: refresh repository identity after SSH translation

benekuehn:
## Summary

- refresh cached remote metadata after resolving SSH host aliases
- preserve fetch URL precedence, falling back to push metadata when the fetch URL cannot be parsed

## Root cause

`repository.Current` translated the fetch and push URLs, but the cached `Host`, `Owner`, and `Repo` fields still described the pre-translation URL. Known-host filtering therefore rejected remotes that used SSH aliases for GitHub.

## Testing

- `go test ./...`
- `go vet ./...`

## Related issues

- Relates to github/gh-stack#45
- Relates to github/gh-stack#337

### DIFF
--- a/pkg/repository/repository.go
+++ b/pkg/repository/repository.go
@@ -5,6 +5,7 @@ package repository
 import (
 	"errors"
 	"fmt"
+	"net/url"
 	"os"
 	"strings"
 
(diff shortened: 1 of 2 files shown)


RishabJaiswal:
Request to merge this urgently

Quite a common usecase for users with multiple github accounts

PR #275, opened 2026-07-30
Add per-host API endpoint overrides and a configurable redirect policy

williammartin:
Related: cli/cli#13717, cli/cli#13991

### Description

GitHub CLI derives REST and GraphQL endpoints from the authenticated GitHub host. Organizations that require API traffic to pass through a corporate gateway currently cannot configure that routing in `gh` or extensions built on go-gh.

This adds a per-host `api_host` setting in `hosts.yml` and an exported `ClientOptions.APIHost` override. Authentication and token selection still use the canonical GitHub host; only the host of endpoints constructed by go-gh changes. Canonical schemes and paths are preserved, including GHES `/api/v3/` and `/api/graphql` paths and `github.localhost` HTTP endpoints.

All three client constructors resolve the override even when callers already supply `Host`, `AuthToken`, and `Transport`. The REST and GraphQL constructors swap the host only after deriving their canonical endpoints. REST absolute URLs supplied by callers or servers remain unchanged. Authorization is permitted only for the canonical domain or one exact configured API host.

The second change here is a `CheckRedirect` field on `ClientOptions`, matching the field of the same name on `http.Client`. `NewHTTPClient` built its client without it, so a caller had no way to control redirect handling. The policy matters because Go's default converts a `DELETE` into a `GET` when it follows a 301, so deleting a repository that has been renamed reports success having deleted nothing. Behaviour is unchanged when `CheckRedirect` is nil.

See all the test harnessing on https://github.com/cli/cli/pull/14104

### Key points

- `api_host` accepts a bare hostname only. It intentionally rejects ports, schemes, and path prefixes, so the mapping is direct and canonical endpoint layout remains authoritative.
- A non-empty `ClientOptions.APIHost` takes precedence over `hosts.yml`; environment-variable configuration is intentionally out of scope.
- Server-provided absolute URLs are not rewritten. A gateway is responsible for rewriting pagination or asset URLs when those requests must remain on its route.
- Config-read failures are treated as no configured API host, preserving existing behavior for callers that fully specify client options. A successfully read but invalid `api_host` still fails client construction.
- `CheckRedirect` is a general field mirroring `http.Client` rather than a narrow "do not follow redirects" flag. go-gh is a library and the policy is the caller's to choose; cli/cli wraps it in a narrow option of its own, where exactly one use case is known.
- Strictly, a caller can honour `api_host` without this field by resolving the API host itself and continuing to build its own client. cli/cli does this in one place, `gh api`, which needs to set `Content-Length` for `--input` uploads and so cannot use the shared request path. That is an escape hatch rather than a pattern: it duplicates resolution and leaves each call site able to forget.

### Notes for reviewers

None

### Authorship and follow-up

Who wrote this:

- [ ] A human wrote it.

Who answers review comments:

### DIFF
--- a/pkg/api/api_host_test.go
+++ b/pkg/api/api_host_test.go
@@ -110,10 +110,11 @@ func newAPIHostTestHarness(t *testing.T, apiHost string) *apiHostTestHarness {
 
 	gateway := httptest.NewTLSServer(http.HandlerFunc(func(w http.ResponseWriter, req *http.Request) {
 		harness.gatewayRequests.record(req)
-		if req.URL.Path == "/redirect-to-third-party" {
+		switch req.URL.Path {
+		case "/redirect-to-third-party":
 			http.Redirect(w, req, "https://"+thirdPartyHost+"/redirected", http.StatusFound)
 			return
-		} else if req.URL.Path == "/redirect-to-subdomain" {
+		case "/redirect-to-subdomain":
 			http.Redirect(w, req, "https://subdomain.gw.example.net/redirected", http.StatusFound)
 			return
 		}
--- a/pkg/api/api_host_test.go
+++ b/pkg/api/api_host_test.go
@@ -19,6 +19,10 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
+// thirdPartyHost is a host that is neither the canonical API host nor the configured
+// api_host, and which must therefore never be sent the auth token.
+const thirdPartyHost = "unrelated.example"
+
 type recordedRequest struct {
 	method        string
 	path          string
@@ -59,9 +63,10 @@ func requireRequest(t *testing.T, recorder *requestRecorder, want recordedReques
 }
 
 type apiHostTestHarness struct {
-	transport         *http.Transport
-	githubAPIRequests requestRecorder
-	gatewayRequests   requestRecorder
+	transport          *http.Transport
+	githubAPIRequests  requestRecorder
+	gatewayRequests    requestRecorder
+	thirdPartyRequests requestRecorder
 }
 
 func newAPIHostTestHarness(t *testing.T, apiHost string) *apiHostTestHarness {
@@ -105,9 +110,22 @@ func newAPIHostTestHarness(t *testing.T, apiHost string) *apiHostTestHarness {
 
 	gateway := httptest.NewTLSServer(http.HandlerFunc(func(w http.ResponseWriter, req *http.Request) {
 		harness.gatewayRequests.record(req)
+		if req.URL.Path == "/redirect-to-third-party" {
+			http.Redirect(w, req, "https://"+thirdPartyHost+"/redirected", http.StatusFound)
+			return
+		}
 		proxy.ServeHTTP(w, req)
 	}))
 	t.Cleanup(gateway.Close)
+
+	// Finally we stand up a server representing an unrelated third party, which must never
+	// receive the auth token, whether it is requested directly or arrived at via a redirect.
+	thirdParty := httptest.NewTLSServer(http.HandlerFunc(func(w http.ResponseWriter, req *http.Request) {
+		harness.thirdPartyRequests.record(req)
+		w.Header().Set(contentType, jsonContentType)
+		_, _ = io.WriteString(w, `{"message":"third party response"}`)
+	}))
+	t.Cleanup(thirdParty.Close)
 
 	// To allow us to use fake domain names that are representative, we use a custom transport
 	// that rewrites the hostnames to point to our test servers.
@@ -118,6 +136,7 @@ func newAPIHostTestHarness(t *testing.T, apiHost string) *apiHostTestHarness {
 		"api.example.ghe.com:443": fakeAddress,
 		"ghes.example.com:443":    fakeAddress,
 		"gw.example.net:443":      gatewayAddress,
+		thirdPartyHost + ":443":   thirdParty.Listener.Addr().String(),
 	}
 	harness.transport = &http.Transport{
 		// We must turn off TLS verification because our test servers use self-signed certs.
@@ -294,6 +313,48 @@ func TestAPIHostRouting(t *testing.T) {
 					rawQuery:      "page=2",
 					host:          apiHost,
 					authorization: "token test-token",
+				})
+			})
+
+			t.Run("does not provide token to an unrelated host", func(t *testing.T) {
+				harness, opts := newConfiguredAPIHostTest(t, tt.host, apiHost)
+				restClient, err := NewRESTClient(opts)
+				require.NoError(t, err)
+
+				var result struct {
+					Message string `json:"message"`
+				}
+				require.NoError(t, restClient.Get("https://"+thirdPartyHost+"/third-party", &result))
+				assert.Equal(t, "third party response", result.Message)
+
+				requireRequest(t, &harness.thirdPartyRequests, recordedRequest{
+					method:        http.MethodGet,
+					path:          "/third-party",
+					host:          thirdPartyHost,
+					authorization: "",
+				})
+			})
+
+			t.Run("does not provide token when redirected off the api_host", func(t *testing.T) {
+				harness, opts := newConfiguredAPIHostTest(t, tt.host, apiHost)
+				httpClient, err := NewHTTPClient(opts)
+				require.NoError(t, err)
+
+				response, err := httpClient.Get("https://" + apiHost + "/redirect-to-third-party")
+				require.NoError(t, err)
+				require.NoError(t, response.Body.Close())
+
+				requireRequest(t, &harness.gatewayRequests, recordedRequest{
+					method:        http.MethodGet,
+					path:          "/redirect-to-third-party",
+					host:          apiHost,
+					authorization: "token test-token",
+				})
+				requireRequest(t, &harness.thirdPartyRequests, recordedRequest{
+					method:        http.MethodGet,
+					path:          "/redirected",
+					host:          thirdPartyHost,
+					authorization: "",
 				})
 			})
 		})
--- a/pkg/api/api_host_test.go
+++ b/pkg/api/api_host_test.go
@@ -1,0 +1,363 @@
+package api
+
+import (
+	"context"
+	"crypto/tls"
+	"fmt"
+	"io"
+	"net"
+	"net/http"
+	"net/http/httptest"
+	"net/http/httputil"
+	"net/url"
+	"strings"
+	"sync"
+	"testing"
+
+	"github.com/cli/go-gh/v2/internal/testutils"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+type recordedRequest struct {
+	method        string
+	path          string
+	rawQuery      string
+	host          string
+	authorization string
+}
+
+type requestRecorder struct {
+	mu       sync.Mutex
+	requests []recordedRequest
+}
+
+func (recorder *requestRecorder) record(req *http.Request) {
+	recorder.mu.Lock()
+	defer recorder.mu.Unlock()
+
+	recorder.requests = append(recorder.requests, recordedRequest{
+		method:        req.Method,
+		path:          req.URL.Path,
+		rawQuery:      req.URL.RawQuery,
+		host:          req.Host,
+		authorization: req.Header.Get("Authorization"),
+	})
+}
+
+func (recorder *requestRecorder) recordedRequests() []recordedRequest {
+	recorder.mu.Lock()
+	defer recorder.mu.Unlock()
+
+	return append([]recordedRequest(nil), recorder.requests...)
+}
+
+func requireRequest(t *testing.T, recorder *requestRecorder, want recordedRequest) {
+	t.Helper()
+
+	require.Contains(t, recorder.recordedRequests(), want)
+}
+
+type apiHostTestHarness struct {
+	transport         *http.Transport
+	githubAPIRequests requestRecorder
+	gatewayRequests   requestRecorder
+}
+
+func newAPIHostTestHarness(t *testing.T, apiHost string) *apiHostTestHarness {
+	t.Helper()
+
+	harness := &apiHostTestHarness{}
+
+	// First we stand up a TLS server that fakes the real GitHub API. It will record requests and respond with canned responses.
+	fakeGitHub := httptest.NewTLSServer(http.HandlerFunc(func(w http.ResponseWriter, req *http.Request) {
+		harness.githubAPIRequests.record(req)
+		w.Header().Set(contentType, jsonContentType)
+		switch req.URL.Path {
+		case "/http-client":
+			_, _ = io.WriteString(w, `{"message":"http client response"}`)
+		case "/direct-api-host":
+			_, _ = io.WriteString(w, `{"message":"direct api host response"}`)
+		case "/repos/cli/example-repository", "/api/v3/repos/cli/example-repository":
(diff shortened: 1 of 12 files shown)


pkg/api/client_options.go
```diff
@@ -104,3 +111,44 @@ func resolveOptions(opts ClientOptions) (ClientOptions, error) {
...
+		u.User != nil ||
+		u.Path != "" ||
+		u.RawQuery != "" ||
+		u.Fragment != "" ||
+		strings.Contains(u.Hostname(), ":") {
+		return false
+	}
+
+	return !strings.Contains(apiHost, ":")
+}
```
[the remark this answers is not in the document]
williammartin: Good catch, and it went further than that: `u.Host != apiHost` already subsumes the `u.User`, `u.Path`, `u.RawQuery` and `u.Fragment` checks too, since any of those make the parsed host differ from the input. Collapsed the whole thing down to a whitespace/colon check plus a single parse, verified behaviour is unchanged across the existing table (scheme, path, query, fragment, userinfo, ports, whitespace) and added IPv6 literal cases, which were the only inputs relying on the `Hostname()` colon check that were not already pinned by a test.

pkg/api/host.go
```diff
@@ -0,0 +1,27 @@
...
+	if err != nil || cfg == nil {
+		return "", false
+	}
+
+	normalizedHost := auth.NormalizeHostname(host)
+	apiHost, err := cfg.Get([]string{hostsKey, normalizedHost, apiHostKey})
+	if err != nil || apiHost == "" {
+		return "", false
+	}
+	return apiHost, true
```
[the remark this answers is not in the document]
williammartin: Agreed, renamed the local to `configuredAPIHost`, which also matches what `resolveAPIHost` already calls the same value.

pkg/api/client_options.go
```diff
@@ -104,3 +129,36 @@ func resolveOptions(opts ClientOptions) (ClientOptions, error) {
...
+		opts.APIHost = configuredAPIHost
+	}
+
+	if !validAPIHost(opts.APIHost) {
+		return ClientOptions{}, fmt.Errorf(
+			`invalid api_host for %s: %q must be a hostname without a scheme or port, for example "api.example.com"`,
+			opts.Host,
+			opts.APIHost,
+		)
+	}
```
babakks: **nitpick:** this error message is about the logic in the `validAPIHost` function, so it's best to move the message there and change the `validAPIHost` signature to return an `error` instead of a boolean. Then this call site will be changed to:

```go
if err := validAPIHost(opts.APIHost); err != nil {
	return ClientOptions{}, fmt.Errorf("invalid api_host for %s: %w", opts.Host, err)
}
```

pkg/api/client_options_test.go
```diff
@@ -150,6 +153,220 @@ func TestOptionsNeedResolution(t *testing.T) {
...
+			host:    "example.ghe.com",
+			apiHost: "https://explicit.example.net",
+			wantErr: `invalid api_host for example.ghe.com: "https://explicit.example.net" must be a hostname without a scheme or port, for example "api.example.com"`,
+		},
+		{
+			name:    "explicit port fails",
+			host:    "example.ghe.com",
+			apiHost: "explicit.example.net:8443",
+			wantErr: `invalid api_host for example.ghe.com: "explicit.example.net:8443" must be a hostname without a scheme or port, for example "api.example.com"`,
+		},
```
babakks, line 241: To further reinforce this, we need the same test cases for the configuration path:


```suggestion
		{
			name:    "invalid explicit option fails",
			host:    "example.ghe.com",
			apiHost: "https://explicit.example.net",
			wantErr: `invalid api_host for example.ghe.com: "https://explicit.example.net" must be a hostname without a scheme or port, for example "api.example.com"`,
		},
		{
			name:    "explicit port fails",
			host:    "example.ghe.com",
			apiHost: "explicit.example.net:8443",
			wantErr: `invalid api_host for example.ghe.com: "explicit.example.net:8443" must be a hostname without a scheme or port, for example "api.example.com"`,
		},
		{
			name:    "invalid configured option fails",
			host:    "example.ghe.com",
			config:  "hosts:\n  example.ghe.com:\n    api_host: https://configured.example.net\n",
			wantErr: `invalid api_host for example.ghe.com: "https://configured.example.net" must be a hostname without a scheme or port, for example "api.example.com"`,
		},
		{
			name:    "configured port fails",
			host:    "example.ghe.com",
			config:  "hosts:\n  example.ghe.com:\n    api_host: configured.example.net:8443\n",
			wantErr: `invalid api_host for example.ghe.com: "configured.example.net:8443" must be a hostname without a scheme or port, for example "api.example.com"`,
		},
```

pkg/api/host_test.go
```diff
@@ -0,0 +1,88 @@
...
+		wantOK bool
+	}{
+		{
+			name: "missing api_host",
+			host: "example.ghe.com",
+			config: `
+hosts:
+  example.ghe.com:
+    oauth_token: token
+`,
```
babakks: Let's use `heredoc.Doc` for all these test cases.

pkg/api/rest_client.go
```diff
@@ -28,22 +28,32 @@ func DefaultRESTClient() (*RESTClient, error) {
 // and unix domain socket are resolved from the gh environment configuration.
```
babakks: We should include the API host among these:

```suggestion
// As part of the configuration a hostname, auth token, default set of headers,
// API host, and unix domain socket are resolved from the gh environment configuration.
```

pkg/api/rest_client.go
```diff
@@ -14,8 +14,8 @@ import (
 // RESTClient wraps methods for the different types of
 // API requests that are supported by the server.
 type RESTClient struct {
-	client *http.Client
-	host   string
+	client   *http.Client
+	endpoint string
```
babakks: This `endpoint` field is actually used as a URL prefix. So, I agree that `host` wasn't a good name, but `endpoint` is not really clear, too. My suggestion is to rename it to `baseURL` or even just `prefix`.

If we do this, we also need to change lots of other semantic references (e.g. in function parameters or tests).

pkg/api/rest_client.go
```diff
@@ -156,11 +166,11 @@ func (c *RESTClient) Put(path string, body io.Reader, resp interface{}) error {
 	return c.Do(http.MethodPut, path, body, resp)
 }
 
-func restURL(hostname string, pathOrURL string) string {
+func restURL(endpoint string, pathOrURL string) string {
 	if strings.HasPrefix(pathOrURL, "https://") || strings.HasPrefix(pathOrURL, "http://") {
 		return pathOrURL
 	}
-	return restPrefix(hostname) + pathOrURL
+	return endpoint + pathOrURL
```
babakks: when `endpoint` is populated with the API host value from config, there's no guarantee it ends with forward slash, so this concat may fail to produce the right path unless the given `pathOrURL` has a leading slash.

Let's ensure there's a slash in the middle:


```suggestion
	if strings.HasSuffix(endpoint, "/") || strings.HasPrefix(pathOrURL, "/") {
		return endpoint + pathOrURL
	}
	return endpoint + "/" + pathOrURL
```

pkg/api/rest_client.go
```diff
@@ -28,22 +28,32 @@ func DefaultRESTClient() (*RESTClient, error) {
...
+
 	client, err := NewHTTPClient(opts)
 	if err != nil {
 		return nil, err
 	}
 
+	endpoint := restPrefix(opts.Host)
+	if opts.APIHost != "" {
+		endpoint = swapHost(endpoint, opts.APIHost)
+	}
```
babakks: Let's add a time saver comment here:


```suggestion
	endpoint := restPrefix(opts.Host)
	if opts.APIHost != "" {
		// The endpoint comes with a trailing slash and swapping the host should preserve it.
		endpoint = swapHost(endpoint, opts.APIHost)
	}
```

pkg/api/graphql_client.go
babakks, line 24: It actually makes more sense as `endpoint` here in the `GraphQLClient` struct as the API endpoint is always the same for GraphQL.

pkg/api/http_client.go
```diff
@@ -50,14 +51,19 @@ func DefaultHTTPClient() (*http.Client, error) {
...
 		opts, err = resolveOptions(opts)
 		if err != nil {
 			return nil, err
 		}
 	}
 
+	opts, err = resolveAPIHost(opts)
+	if err != nil {
+		return nil, err
+	}
```
babakks, line 65: **remark/suggestion:** Both `NewRESTClient` and `NewGraphQLClient` already call `resolveAPIHost` on the `opts` value. The `resolveAPIHost` function is carefully implemented so that the second call doesn't try reading the config, but the validation part happens twice (obviously, the second time is always a pass).

I think this implies we should take out the validation from `resolveAPIHost` and do it here as a separate/explicit call. 🤔

pkg/api/api_host_test.go
```diff
@@ -0,0 +1,424 @@
...
+
+				requireRequest(t, &harness.thirdPartyRequests, recordedRequest{
+					method:        http.MethodGet,
+					path:          "/third-party",
+					host:          thirdPartyHost,
+					authorization: "",
+				})
+			})
+
+			t.Run("does not provide token when redirected off the api_host", func(t *testing.T) {
```
babakks, line 343: We need another test case to confirm the token is not sent when redirecting to subdomains of the API host. Here's the diff to add that (and it's passing):

```patch
diff --git a/pkg/api/api_host_test.go b/pkg/api/api_host_test.go
index 3b481b3..5687ede 100644
--- a/pkg/api/api_host_test.go
+++ b/pkg/api/api_host_test.go
@@ -113,6 +113,9 @@ func newAPIHostTestHarness(t *testing.T, apiHost string) *apiHostTestHarness {
 		if req.URL.Path == "/redirect-to-third-party" {
 			http.Redirect(w, req, "https://"+thirdPartyHost+"/redirected", http.StatusFound)
 			return
+		} else if req.URL.Path == "/redirect-to-subdomain" {
+			http.Redirect(w, req, "https://subdomain.gw.example.net/redirected", http.StatusFound)
+			return
 		}
 		proxy.ServeHTTP(w, req)
 	}))
@@ -132,11 +135,12 @@ func newAPIHostTestHarness(t *testing.T, apiHost string) *apiHostTestHarness {
 	fakeAddress := fakeGitHub.Listener.Addr().String()
 	gatewayAddress := gateway.Listener.Addr().String()
 	dialMap := map[string]string{
-		"api.github.com:443":      fakeAddress,
-		"api.example.ghe.com:443": fakeAddress,
-		"ghes.example.com:443":    fakeAddress,
-		"gw.example.net:443":      gatewayAddress,
-		thirdPartyHost + ":443":   thirdParty.Listener.Addr().String(),
+		"api.github.com:443":           fakeAddress,
+		"api.example.ghe.com:443":      fakeAddress,
+		"ghes.example.com:443":         fakeAddress,
+		"gw.example.net:443":           gatewayAddress,
+		"subdomain.gw.example.net:443": gatewayAddress,
+		thirdPartyHost + ":443":        thirdParty.Listener.Addr().String(),
 	}
 	harness.transport = &http.Transport{
 		// We must turn off TLS verification because our test servers use self-signed certs.
@@ -357,6 +361,28 @@ func TestAPIHostRouting(t *testing.T) {
 					authorization: "",
 				})
 			})
+
+			t.Run("does not provide token when redirected to a subdomain of the api_host", func(t *testing.T) {
+				harness, opts := newConfiguredAPIHostTest(t, tt.host, apiHost)
+				httpClient, err := NewHTTPClient(opts)
+				require.NoError(t, err)
+
+				response, err := httpClient.Get("https://" + apiHost + "/redirect-to-subdomain")
+				require.NoError(t, err)
+				require.NoError(t, response.Body.Close())
+
+				requireRequest(t, &harness.gatewayRequests, recordedRequest{
+					method:        http.MethodGet,
+					path:          "/redirect-to-subdomain",
+					host:          apiHost,
+					authorization: "token test-token",
+				})
+				requireRequest(t, &harness.gatewayRequests, recordedRequest{
+					method: http.MethodGet,
+					path:   "/redirected",
+					host:   "subdomain." + apiHost,
+				})
+			})
 		})
 	}
 
```

pkg/api/client_options.go
```diff
@@ -25,6 +40,16 @@ type ClientOptions struct {
...
 
+	// CheckRedirect specifies the policy for handling redirects, matching the
+	// field of the same name on http.Client. If nil, the default policy of
+	// following up to 10 redirects is used.
+	//
+	// This matters for requests where following a redirect silently changes
+	// the meaning of the request. Go's default policy converts a DELETE into a
+	// GET when it follows a 301, so a caller deleting a renamed resource can
+	// receive a success response having deleted nothing.
+	CheckRedirect func(*http.Request, []*http.Request) error
```
babakks, line 55: **nitpick:** to be clear it's not just `DELETE`:

```suggestion
	// This matters for requests where following a redirect silently changes
	// the meaning of the request. For example, Go's default policy converts a DELETE into a
	// GET when it follows a 301, so a caller deleting a renamed resource can
	// receive a success response having deleted nothing.
	CheckRedirect func(*http.Request, []*http.Request) error
```

babakks (changes requested):
Thanks for putting this together, @williammartin! :pray:

I reviewed the whole change set, and overall the design looks solid to me. The separation between authentication (keyed on `Host`) and routing (keyed on `APIHost`), the decision to never rewrite absolute URLs, and the exact-match token boundary for the override host all feel well reasoned. I didn't find any correctness or security issues with the system itself.

I left a handful of inline comments, mostly naming, godoc, and test-coverage nitpicks, plus a couple of small refactor suggestions.

The one thing I'd like to see landed is a test asserting that the token is withheld when a request is redirected to a subdomain of the configured `api_host`. Right now the suite proves that canonical-host subdomains receive the token and that unrelated third parties don't, but the "exact match, no parent-domain pass" guarantee for `api_host` is only covered indirectly. I've added an inline suggestion with a passing patch that wires this case in, so it should be a quick add.

pkg/api/graphql_client.go
babakks: Fixed in c414411

pkg/api/rest_client.go
babakks: Fixed in ab5eab2

pkg/api/client_options_test.go
babakks: Fixed in 01a48cf

pkg/api/rest_client.go
babakks: Fixed in c414411

pkg/api/rest_client.go
babakks: Fixed in ab5eab2

pkg/api/api_host_test.go
babakks: Fixed in 6d1c34d

pkg/api/client_options.go
babakks: Fixed in ab5eab2

pkg/api/host_test.go
babakks: Fixed in ec70c9c

pkg/api/client_options.go
babakks: After reconsidering this one, I've decided to skip it for now.

pkg/api/http_client.go
babakks: After reconsidering this one, I've decided to skip it for now.

babakks (approved):
LGTM

PR #277, opened 2026-08-07
Let clients configure a redirect policy

williammartin:
`NewHTTPClient` builds an `http.Client` from `ClientOptions` but never sets
`CheckRedirect`, so a caller has no way to control redirect handling and always
gets Go's default policy.

That default is not always what a caller wants. Go converts a `DELETE` into a
`GET` when it follows a 301, so a client deleting a resource that has been
renamed receives a success response having deleted nothing.

This adds a `CheckRedirect` field to `ClientOptions` and threads it through, in
the same shape as `http.Client`'s field of the same name. Behaviour is unchanged
when it is nil.

### Why now

cli/cli is migrating its remaining raw `httpClient.Do` call sites onto its
`api.Client`, so that all API traffic honours the per-host `api_host` config
added in #275. `gh repo delete` was the one site that could not migrate: it
needs this policy, and because go-gh discarded it, the command had to build its
own client and hold an absolute `api.github.com` URL, which meant it silently
ignored `api_host`.

With this change it migrates, and an end-to-end test that routes every request
through a gateway with `api.github.com` blackholed passes in full. That test is
what surfaced this: `gh repo delete` was the last command still reaching
GitHub directly.

Together with #275, this is everything go-gh needs for that work. Draft because
#275 should land first.

### Testing

`TestNewHTTPClientCheckRedirect` covers both the default policy and a custom one
that stops at the redirect. It uses the package's existing `tripper` stub;
redirect policy lives on `http.Client` above the transport, so a stubbed
transport still exercises the real policy.

Verified the test detects the bug: reverting the one-line change to
`NewHTTPClient` fails it with `expected: 301, actual: 204`.

### DIFF
--- a/pkg/api/client_options.go
+++ b/pkg/api/client_options.go
@@ -25,6 +25,16 @@ type ClientOptions struct {
 	// Default is 24 hours.
 	CacheTTL time.Duration
 
+	// CheckRedirect specifies the policy for handling redirects, matching the
+	// field of the same name on http.Client. If nil, the default policy of
+	// following up to 10 redirects is used.
+	//
+	// This matters for requests where following a redirect silently changes
+	// the meaning of the request. Go's default policy converts a DELETE into a
+	// GET when it follows a 301, so a caller deleting a renamed resource can
+	// receive a success response having deleted nothing.
+	CheckRedirect func(*http.Request, []*http.Request) error
+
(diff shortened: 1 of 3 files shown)


williammartin:
Folded into #275 so there is a single PR carrying everything go-gh needs for the cli/cli api_host migration. The change itself is unmodified.

PR #281, opened 2026-08-13
Upgrade glamour to fix hyperlinks (OSC8)

### LINKED ISSUE #188: Pick up new version of Glamour to avoid breaking hyperlinks
### Describe the feature or problem you’d like to solve

When rendering markdown, hyperlinks that wrap or are cut off get broken, such that a line with "https://git" and another with "hub.com/cli/cli" will have at least one link that doesn't work (or, worse, goes somewhere NSFW, etc.) and another line that probably won't be recognized as a link.

Fortunately, this issue was fixed in Glamour: https://github.com/charmbracelet/glamour/pull/411

Once they cut a new release, it would be great to pick that up.
 
While I thought I opened or commented on a bug on this, I can't seem to find it. Apologies if it's dup'd.

### Proposed solution

Hyperlinks in rendered markdown aren't broken if they wrap lines or are otherwise cut off.

### Additional context

This is unrelated to years-long discussions on hyperlinks (OSC8) printed directly by `gh`, though it might be worth revisiting that in 2025. At least as of 2022, the main concern seemed to be Mac's default Terminal app...though, doesn't everyone replace that with iTerm2 or something these days. :wink: (I do appreciate that problem, though).

But since you use Glamour anyway for markdown rendering, it would be good to pick this new version up to fix broken hyperlinks when it renders them.

heaths:
Fixes #188


heaths:
This is stacked atop #280 so this either needs to be split into a separate PR atop `trunk` or turned into a stack / that PR merged first. I'm building up my own replacement module with fixes I need for extensions we use in workflows and custom agent instructions.

heaths:
This is part of a stack but not showing here because stacks aren't enabled for this repo it seems:

```text
Syncing PRs ...
✓ PR #280 (issue279) — Open
✓ PR #281 (issue188) — Open
⚠ Stacked PRs are not enabled for this repository

✓ Branches synced
  Stacked on trunk (450618a)
```

I can split these later if needs be, but stacks at least make it easier to `rebase --onto` (automatic vs. [manual](https://heaths.dev/tips/2019/11/11/rebasing-commits-on-one-topic-branch-onto-another-branch.html)).

BagToad:
I would prefer this dependency bump on its own and not stacked with the unrelated change please. If you have the time; proof that it doesn't break tests in cli/cli would also be an asset to decrease the review burden. If it does break tests,   some investigation into why is in order as well.

I don't understand the "replacement module" comment, sorry. If that's something you meant for us to action, can you expand on that?

heaths:
> I would prefer this dependency bump on its own and not stacked with the unrelated change please.

@BagToad if you're open to taking it, I can unstack them. I mainly did that so I could simply build against the head of the stack against my fork. Less for the stack and more for the natural `HEAD` so I can build against all changes.

I'm making a similar change to the cli right now for cli/cli#3718. There were a few expected `\e[0m` -> `\e[m` conversions (effectively equivalent) but also a couple other changes with the newer version of lipgloss that you already used in some places (your `go.mod` for `gh` uses both the v2 and v0 versions of lipgloss.

But how would this change impact the cli directly? I see you are using `x/markdown`. I should probably split this PR out and use a replacement module (literally "replace" in the `go.mod` e.g., <https://github.com/heaths/gh-minimize/blob/710a1ab366b560af52e84846379a81df05ed00b5/go.mod#L14>) there as well.

heaths:
The lint issue is that `golangci-lint` timed out or otherwise couldn't connect to the schema. Odd thing for it to fail on; however, I can't rerun the failed check myself. I imagine it'd pass on a subsequent run. The file exists—I just checked in.

heaths:
Since I can't rerun the failed test I simply amended the commit and did a force push.

PR #288, opened 2026-09-02
Add guarded release workflow

williammartin:
## Summary

- Add a maintainer-facing `script/release` command that validates a requested version and dispatches the release workflow from `trunk`.
- Add reusable release validation covering version format, duplicate tags/releases, release ordering, commit ancestry, and required checks.
- Publish releases from an exact, revalidated `trunk` commit after approval through the `release` environment.
- Restrict release publication to workflow-dispatch runs with a write-scoped job token.

## Repository setup

Configure required reviewers for the `release` environment before using this workflow.

## Validation

- `script/validate-release v2.14.0`
- ShellCheck
- zizmor

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,58 @@
+name: Release
+run-name: Release ${{ inputs.version }}
+
+on:
+  workflow_dispatch:
+    inputs:
+      version:
+        description: Version to release (v2.MINOR.PATCH)
+        required: true
+        type: string
+
+concurrency:
+  group: release
+  cancel-in-progress: false
+
+permissions:
+  checks: read # Verify required checks on the release commit.
+  contents: read # Check out and inspect the release commit.
+
+jobs:
+  validate:
+    name: Validate release
+    runs-on: ubuntu-latest
+    timeout-minutes: 5
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+        with:
+          fetch-depth: 0
+          persist-credentials: false
+
+      - name: Validate release
+        env:
+          GH_TOKEN: ${{ github.token }}
+          VERSION: ${{ inputs.version }}
+        run: script/validate-release "$VERSION" "$GITHUB_SHA"
+
+  release:
+    name: Publish release
+    needs: validate
+    runs-on: ubuntu-latest
+    timeout-minutes: 5
+    environment: release
+    permissions:
+      checks: read # Reverify required checks after environment approval.
+      contents: write # Create the release and its tag.
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+        with:
+          fetch-depth: 0
+          persist-credentials: false
+
+      - name: Publish release
+        env:
+          GH_TOKEN: ${{ github.token }}
+          VERSION: ${{ inputs.version }}
+        run: script/publish-release "$VERSION" "$GITHUB_SHA"
--- a/script/publish-release
+++ b/script/publish-release
@@ -1,11 +1,18 @@
 #!/usr/bin/env bash
 
+if [ -z "${BASH_VERSION:-}" ]; then
+	printf 'error: this script must be run with Bash\n' >&2
+	exit 1
+fi
+
 set -euo pipefail
 
 SCRIPT_DIR=$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)
 readonly SCRIPT_DIR
 readonly REPOSITORY="cli/go-gh"
+readonly BRANCH="trunk"
 readonly RELEASE_MAJOR="2"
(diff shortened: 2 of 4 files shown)


.github/workflows/release.yml
```diff
@@ -0,0 +1,58 @@
...
+          GH_TOKEN: ${{ github.token }}
+          VERSION: ${{ inputs.version }}
+        run: script/validate-release "$VERSION" "$GITHUB_SHA"
+
+  release:
+    name: Publish release
+    needs: validate
+    runs-on: ubuntu-latest
+    timeout-minutes: 5
+    environment: release
```
williammartin, line 43: This requires approval from an environment `release` but note that it **doesn't** block any user with write from just creating a tag. What we need to do is create a rule set that prevents tag creation, then register a specific github app that can bypass it and mint credentials for that app on release.

script/validate-release
```diff
@@ -0,0 +1,139 @@
...
+latest_version=$(gh release view \
+	--repo "$REPOSITORY" \
+	--json tagName \
+	--jq .tagName)
+if [[ ! $latest_version =~ ^v${RELEASE_MAJOR}\.(0|[1-9][0-9]*)\.(0|[1-9][0-9]*)$ ]]; then
+	fail "latest release has an unexpected version: $latest_version"
+fi
+
+latest_minor=${BASH_REMATCH[1]}
+latest_patch=${BASH_REMATCH[2]}
```
babakks: **nitpick:** this and the string split (`IFS`) only work in bash, so if a maintainer accidentally runs this via like `zsh script/validate-release` it may not work properly.

I suggest adding a simple check like this at the top of the script, or extract the regex capture groups in a different way:

```sh
if [[ -z ${BASH_VERSION:-} ]]; then
    printf 'Error: this script must be run with Bash.\n' >&2
    exit 1
fi
```

script/publish-release
```diff
@@ -0,0 +1,69 @@
...
+	usage >&2
+	exit 1
+}
+
+version=$1
+target_sha=$2
+
+[[ ${GITHUB_ACTIONS:-} == "true" &&
+	${GITHUB_EVENT_NAME:-} == "workflow_dispatch" ]] ||
+	fail "releases may only be published by the release workflow"
```
babakks, line 47: Nice!

babakks (approved):
LGTM, with a just a quick suggestion.

script/validate-release
```diff
@@ -0,0 +1,139 @@
...
+	exit 1
+}
+
+usage() {
+	cat <<EOF
+Usage: script/validate-release v${RELEASE_MAJOR}.MINOR.PATCH [TARGET_SHA]
+
+Validate a go-gh release against the current origin/${BRANCH}.
+TARGET_SHA defaults to the current origin/${BRANCH} commit.
+EOF
```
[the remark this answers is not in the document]
williammartin: Addressed in e89e362. The help now states that `TARGET_SHA`, when provided, must match the current `origin/trunk` commit.

script/publish-release
```diff
@@ -0,0 +1,69 @@
...
+
```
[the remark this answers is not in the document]
williammartin: Addressed in e89e362. `publish-release` now requires `GITHUB_WORKFLOW_REF` to identify `.github/workflows/release.yml` on `trunk`, in addition to checking Actions and `workflow_dispatch`.

script/validate-release
williammartin: Addressed in e89e362. All three release scripts now fail immediately with a clear error when explicitly invoked through a non-Bash shell.

script/publish-release
williammartin: Thanks!

script/publish-release
```diff
@@ -0,0 +1,69 @@
...
+actual_repository=$(gh repo view --json nameWithOwner --jq .nameWithOwner)
+[[ $actual_repository == "$REPOSITORY" ]] ||
+	fail "expected repository $REPOSITORY, found $actual_repository"
+
+"$SCRIPT_DIR/validate-release" "$version" "$target_sha"
+
+can_push=$(gh api "repos/$REPOSITORY" --jq '.permissions.push')
+[[ $can_push == "true" ]] ||
+	fail "the GitHub token cannot create releases in $REPOSITORY"
+
```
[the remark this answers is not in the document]
williammartin: Addressed in e89e362. I removed the misleading `.permissions.push` preflight; `gh release create` is now the authoritative permission check and surfaces any write-permission failure directly.

babakks (approved):

script/release
```diff
@@ -0,0 +1,75 @@
...
+
+actual_repository=$(gh repo view --json nameWithOwner --jq .nameWithOwner)
+[[ $actual_repository == "$REPOSITORY" ]] ||
+	fail "expected repository $REPOSITORY, found $actual_repository"
+
+"$SCRIPT_DIR/validate-release" "$version"
+
+can_push=$(gh api "repos/$REPOSITORY" --jq '.permissions.push')
+[[ $can_push == "true" ]] ||
+	fail "your GitHub account cannot create releases in $REPOSITORY"
```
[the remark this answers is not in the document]
williammartin: Addressed in ab443ff. I removed the unrelated push-permission preflight; `gh workflow run` now provides the authoritative dispatch authorization check.

PR #289, opened 2026-09-02
Trust canonical host alongside API host

williammartin:
### Description

`api_host` is a routing feature, not a security boundary that should prevent the canonical host from receiving its token.

### Authorship and follow-up

Who wrote this:

- [ ] A human wrote it.

Who answers review comments:


babakks (approved):
LGTM

PR #290, opened 2026-09-07
fix(repository): honor resolved remote

### LINKED ISSUE #286: repository.Current() ignores gh-resolved and selects the wrong repo when a remote is named "github"
`repository.Current()` selects a repository by ranking remotes on their **name** alone, and never consults the `gh-resolved` git config value that `gh repo set-default` writes — even though `Remotes()` reads that value into `Remote.Resolved` immediately beforehand. In a repository where the fork is on a remote named `github` and the parent is on `origin`, `Current()` returns the fork, contradicting both the user's explicit `gh repo set-default` choice and `gh`'s own answer for the same repository.

### Current behaviour

`Remotes()` (`internal/git/remote.go:43`) calls `setResolvedRemotes(remotes)` at line 49, populating `Remote.Resolved` (field at line 18, set at line 144) from `git config --get-regexp ^remote\..*\.gh-resolved$`. Line 50 then sorts by `remoteNameSortScore` (line 30): `upstream`=3, `github`=2, `origin`=1, everything else 0. `Current()` (`pkg/repository/repository.go:119`) takes `filteredRemotes[0]` at line 152. `Resolved` appears zero times in the body of `Current()`.

So the value is parsed and then discarded.

### Reproduction

```bash
git init repro && cd repro
git remote add origin git@github.com:parent-org/example.git
git remote add github git@github.com:my-user/example.git
git config remote.origin.gh-resolved base    # what `gh repo set-default` writes
```

```go
package main

import (
	"fmt"
	"github.com/cli/go-gh/v2/pkg/repository"
)

func main() {
	r, err := repository.Current()
	fmt.Printf("Current() -> %s/%s (host %s) err=%v\n", r.Owner, r.Name, r.Host, err)
}
```

Result with go-gh v2.13.0:

```
Current() -> my-user/example (host github.com) err=<nil>
```

Expected `parent-org/example`. `gh repo set-default --view` in the same directory correctly prints `parent-org/example`.

Removing the `gh-resolved` setting produces **identical** output, confirming the value has no effect on the result. `GH_REPO=parent-org/example` does override correctly.

### Expected behaviour

`Current()` should prefer a remote whose `Resolved` value is set, ahead of the name-based ranking. `cli/cli` already has exactly this in `context/remote.go:51`:

```go
func (r Remotes) ResolvedRemote() (*Remote, error) {
	for _, rr := range r {
		if rr.Resolved != "" {
```

go-gh populates the same field but has no equivalent, so the two libraries disagree about the same repository.

The `gh-resolved` support was added in 9cb500d1 ("Add support for selected default remote", 2021-09-27), which suggests wiring it into `Current()` was intended and simply missed.

### Impact

Consumers silently operate on the wrong repository, with no error. Surfaced via `gh-dash`, whose smart filtering injects `repo:<owner>/<name>` from `Current()` and so shows an empty PR list for a fork with no PRs.

### Environment

- go-gh v2.13.0 (and current `trunk` — code unchanged)
- gh 2.96.0, go 1.25.12, Linux

### Prior search

Searched go-gh and cli/cli issues and PRs, all states, for `gh-resolved`, `repository.Current`, remote sorting/precedence, and fork-vs-upstream selection. Closest existing items are #274 (SSH-alias translation) and cli/cli#13027 (proxy URLs failing host detection); neither covers this.

ryux1:
Closes #286

### Description

`repository.Current()` ranked Git remotes by name but ignored the `gh-resolved` configuration written by `gh repo set-default`. This could cause consumers to operate on a fork or another higher-ranked remote instead of the repository explicitly selected by the user.

This change prefers the first resolved remote after the existing host filtering and remote-name ordering. A resolution of `base` selects that remote directly; an explicit repository resolution selects the named owner and repository on the resolved remote's host. When no remote is resolved, the existing name-based fallback remains unchanged.

I created a Go extension that prints the repository returned by `repository.Current()`, then configured a higher-ranked `github` fork remote and marked the `origin` parent remote as the default repository.

Terminal demo showing go-gh v2.16.0 selecting my-user/example before the change and PR 290 selecting parent-org/example after the change

#### Setup

```bash
DEMO_ROOT=$(mktemp -d)
cd "$DEMO_ROOT"
gh extension create --precompiled=go current-repo-demo
cd gh-current-repo-demo

cat > main.go <<'EOF'
package main

import (
	"fmt"
	"os"

	"github.com/cli/go-gh/v2/pkg/repository"
)

func main() {
	current, err := repository.Current()
	if err != nil {
		fmt.Fprintln(os.Stderr, err)
		os.Exit(1)
	}
	fmt.Printf("%s/%s\n", current.Owner, current.Name)
}
EOF

git remote add origin git@github.com:parent-org/example.git
git remote add github git@github.com:my-user/example.git
git config remote.origin.gh-resolved base
```

#### Before

I built the extension against the latest released go-gh version:

```bash
go get github.com/cli/go-gh/v2@v2.16.0
go build -o gh-current-repo-demo .
GH_TOKEN=token ./gh-current-repo-demo
```

```text
my-user/example
```

The extension ignored the resolved `origin` remote and selected the higher-ranked remote named `github`.

#### After

I rebuilt the same extension against this PR:

```bash
git clone https://github.com/cli/go-gh.git "$DEMO_ROOT/go-gh-pr"
git -C "$DEMO_ROOT/go-gh-pr" fetch origin pull/290/head:pr-290
git -C "$DEMO_ROOT/go-gh-pr" checkout pr-290

go mod edit -replace github.com/cli/go-gh/v2="$DEMO_ROOT/go-gh-pr"
go mod tidy
go build -o gh-current-repo-demo .
GH_TOKEN=token ./gh-current-repo-demo
```

```text
parent-org/example
```

The extension respected the repository selected by `gh repo set-default`.

### Key points

- The behavior follows `cli/cli`'s `context.ResolvedRemotes.BaseRepo` resolution order.
- Resolved remotes are considered only after filtering to known GitHub hosts.
- Existing `upstream`, `github`, `origin`, then other-remote precedence remains the fallback.

### Notes for reviewers

Start with `pkg/repository/repository.go` in `Current()`, then review the scenarios in `pkg/repository/repository_test.go`.

Issue #286 contains the original reproduction and explains the behavior difference from `cli/cli`.

### Authorship and follow-up

Who wrote this:

- [ ] A human wrote it.

Who answers review comments:


williammartin (approved):
Thanks @ryux1, I made some changes on your branch.

PR #292, opened 2026-09-09
Return HTTPError from GraphQL query and mutate calls

sidsri14:
Related to cli/cli#8434: `gh pr status` (and other unauthenticated GraphQL callers) reports a raw string like

```
non-200 OK status code: 401 Unauthorized body: "..."
```

instead of a structured, formatted error (`HTTP 401: Bad credentials ...` plus the `gh auth login` hint).

## Root cause

`gh`'s GraphQL calls that go through the named query/mutate path end up in the underlying GraphQL client (`cli/shurcooL-graphql`), which formats any non-2xx HTTP response as a plain `fmt.Errorf("non-200 OK status code: %v body: %q", ...)`. That error carries no status code, headers, or body, so downstream code (`cli/api` `handleResponse`, `gh`'s top-level error handling) has nothing to work with and the ugly raw DNS-style message leaks to the user.

The `Do`/`DoWithContext` path already converts non-2xx responses into `*api.HTTPError` via `HandleHTTPError`; only `Query`/`Mutate` (the `QueryNamed`/`MutateNamed` path) leak the raw error.

## Change

- Add an internal `httpErrorRoundTripper` that converts responses outside the 2xx range into an `*HTTPError` (via the existing `HandleHTTPError`), so the GraphQL client never sees the raw non-2xx response.
- Apply it only to the inner GraphQL HTTP client used for named queries/mutations, leaving the `Do` path and its tests untouched.
- In `QueryWithContext`/`MutateWithContext`, unwrap the `*url.Error` that `net/http` wraps around round-trip failures (`errors.As`) and return the bare `*HTTPError`, consistent with the `Do` path.

Result for an unauthenticated caller is now:

```
HTTP 401: Bad credentials (https://api.github.com/graphql)
Try authenticating with: gh auth login
```

## Tests

- `TestGraphQLClientQueryHTTPError`: 401 -> `HTTPError{StatusCode: 401, Message: "Bad credentials"}`
- `TestGraphQLClientMutateHTTPError`: 404 -> `HTTPError{StatusCode: 404, Message: "Not Found"}`
- `TestGraphQLClientDoHTTPError`: 403 -> `HTTPError{StatusCode: 403, Message: "API rate limit exceeded"}` (guards the untouched `Do` path)
- Existing GraphQL client tests pass unchanged; full `go test ./...` and `go vet ./...` are green.

Note: existing `TestGraphQLClientDo` HTTP-failure cases (which assert `HTTP 404 (https://api.github.com/graphql)` etc.) exercised this behavior before and continue to do so via the `Do` path.

---

*Health disclosure: this pull request was implemented with the assistance of an AI coding tool (opencode). The change was reviewed and verified manually, including running the full test suite and an end-to-end reproduction with a locally built `gh` binary.*


sidsri14:
Thanks for the close - I see go-gh prefers the approach to be agreed in an issue first. Opened cli/go-gh#293 with the proposal (root cause, approach, ready-to-submit implementation). Happy to re-issue the PR once you've reviewed there.

PR #301, opened 2026-09-21
fix(api): evict corrupt cache entries

efegokdemir:
Fixes #300.

## Summary

The file-backed API response cache could serve corrupted or legacy entries indefinitely. This caused callers such as `gh pr list` to receive malformed cached JSON until the user manually cleared the cache.

## Changes

- Write a SHA-256 sidecar for each newly published cache entry.
- Verify the sidecar before parsing and serving cached responses.
- Evict entries with missing or mismatched checksums, and entries that fail HTTP response parsing, so the request transparently refreshes them.
- Keep publication failure cleanup atomic for the response and checksum files.
- Add regression tests for corrupted bodies, legacy entries without checksums, and publication failures.

## Testing

- `go test ./pkg/api` ✅
- `go test ./...` ✅
- `go test -race ./pkg/api` ✅
- `go vet ./...` ✅
- `gofmt -w pkg/api/cache.go pkg/api/cache_test.go` ✅
- `git diff --check` ✅

## Notes

The checksum sidecar intentionally makes entries written by older versions cache misses; they are refreshed once and then become verifiable. `golangci-lint` was not installed in the local environment.


efegokdemir:
Closing this duplicate PR after discovering the existing upstream fix in #300. No action is needed from maintainers.

PR #303, opened 2026-09-21
fix(api): log octocat stream response bodies

efegokdemir:
## Summary

Fix verbose HTTP logging so successful `application/octocat-stream` responses, such as GitHub's `/octocat` endpoint, include their response body.

## Changes

- Treat `application/octocat-stream` as an inspectable MIME type.
- Preserve support for MIME parameters such as a charset.
- Add table-driven regression coverage for octocat-stream and unrelated binary content.

## Testing

- `go test ./...`
- `go vet ./...`
- `git diff --check`

All checks pass.

## Notes

The MIME check uses the existing prefix-based handling, matching the behavior of the other supported content types. This contribution was prepared with AI assistance and manually reviewed.


efegokdemir:
Closing as duplicate of the existing upstream PR #211 for issue #210. No further action is needed here.

